PR #28, opened 2022-11-15
First pass at OAK termset similarity implementation

caufieldjh:


caufieldjh:
FYI the api call function I put in `semsim.py` is like six different kinds of broken

PR #151, opened 2023-04-28
Incorporate new Monarch API node page endpoints into frontend, other fixes

glass-ships:
frontend (roughly in diff order):

- change node page slug to `/node` instead of `/:category`
- update node/search/etc e2e test content
- add `category_label` to fixture data ahead of api incorporating it
- update node page mock handler 
- remove old node and search fixtures
- remove gene api code! (gene symbol and etc should be included in KG. genome not being used atm.)
- remove unused api code, including old (biolink) endpoints that have been merged into single entity endpoint 
- update model.ts
- make use of category_label instead of category for category display and matching (category-specific UI)
- rename relation to predicate
- rename vars related to useQuery func for clarity
- rename faceting/filtering vars for clarity
- move dropdown select show counts prop to computed within component
- update node page sections to use new endpoints. overview, details, hierarchy, associations count
- remove sections from node page that are not ready yet, i.e. associations, evidence, breadcrumbs. they can always be restored from git history later!

PR #153, opened 2023-05-03
API support for association counts with labels

kevinschaper:
- Support for association_counts with labels in the node page response
- formatting
- Swap entity fixture so that it's a disease with inheritance populated and set the neighborhood associations query to get direct only
- Update api fixture tests to go with updated fixture
- A few more content checks on the node page api fixture test
- bump version


glass-ships (approved):

kevinschaper:
Closing, this branch is merged into #151, and they should be able to move in together.

PR #155, opened 2023-05-05
Vite fixes

vincerubinetti:
- format workflow yaml files with prettier
- fix deploy-frontend workflow (needed /frontend working dir)
- fix phenogrid imports and types. this is likely what was preventing playwright (initially) and netlify deploy preview from working in the last pr.
- revert playwright server command to build prod app on CI. this would've caught phenogrid issue, which only happened when building and previewing prod app.
- fix minor tab title bug
- rework icon component to be a little cleaner
- format mkdocs config yaml with prettier

### DIFF
--- a/.github/workflows/build-image.yaml
+++ b/.github/workflows/build-image.yaml
@@ -1,14 +1,15 @@
 name: Docker Build and Push to GCR
-on: 
+on:
   push:
-    branches: [ main ]
+    branches:
+      - main
 
-# This job uses RafikFarhad's GitHub action to build and 
+# This job uses RafikFarhad's GitHub action to build and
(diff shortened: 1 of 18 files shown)


vincerubinetti:
Note that for now, I manually edited the `gh-pages` branch and moved the previously generated docs into a `/docs` subfolder. So the api docs are currently at, and will remain at:
https://monarch-app.monarchinitiative.org/docs/

But, when I merged the last PR, [the build docs workflow failed](https://github.com/monarch-initiative/monarch-app/actions/runs/4887452756) with a bunch of errors that I don't understand. @kevinschaper might need to take a look, or we can figure it out later.

kevinschaper (approved):

PR #162, opened 2023-05-17
Supply a single associations endpoint in favor of association/all etc

kevinschaper:
Looking to make `api/associations?` show up in docs and work without a trailing slash before the question mark. Also hoping to continue to support the previous `association/` endopints while hiding them from the docs


kevinschaper:
Glass took care of this in another PR

PR #163, opened 2023-05-18
Initial backend implementation of association-table api endpoint

kevinschaper:
- Initial backend support for association table api endpoint
- update to latest monarch-py


glass-ships (commented):
are we gonna need to update the linkml version once we get the fixed release out?

kevinschaper:
@glass-ships I do wonder if the test error is because we need a newer linkml or should be pointing to a branch

kevinschaper:
Association types will get replaced with biolink categories soon, which will be a bit of a refactor, but this is still good to bring in

glass-ships (approved):

PR #175, opened 2023-06-13
Fix silly off-by-one constraints error on offset

kevinschaper:
I don't know that we made an issue for this, but this fixes offset having to be greater than equal to 1


glass-ships (approved):
this has been bugging me!! good fix 10/10

PR #182, opened 2023-06-22
implement new monarch-py schema

glass-ships:
- Combines the two schema files
- regenerate models
- simplifies entity/node call in backend


vincerubinetti (approved):
Looks good, thanks for updating the frontend. Maybe we should update the frontend fixture data here too? `node-associations.json` maybe. `association-evidence.json` and `phenotype-explorer-*.json` are in the future and can probably wait.

glass-ships:
> Looks good, thanks for updating the frontend. Maybe we should update the frontend fixture data here too? `node-associations.json` maybe. `association-evidence.json` and `phenotype-explorer-*.json` are in the future and can probably wait.

yeah, i'm actually working on a PR now that introduces a `make` target to generate new fixtures from the data (see #183)
frontend is good, i'm just trying to workout how i'll do the same for the backend as well since pytest fixtures are a little weirder than straight json

PR #183, opened 2023-06-22
Generate fixtures

### LINKED ISSUE #172: Automate generation of frontend fixtures
the json fixtures may end up needing to get changed often, so we may want a way to pass specific queries monarch-py / the current API and pipe that output to json files. or something

glass-ships:
### Related issues

- Closes #172 

Currently only generates:
- node
- node-associations
- histopheno
- search
- autocomplete

We can implement more as we go along, but these felt like the important ones


glass-ships:
Running backend tests successfully (including importing fixtures) only works when run from the top level dir in repo, so I've updated the github action so that it runs. 

frontend e2e test failing though, not sure why

kevinschaper (approved):

PR #189, opened 2023-06-30
Moving monarch-py into monarch-app

glass-ships:
- Merge monarch-py and monarch-api into backend of monarch-app
- Combine and update documentation
- api is now an optional extra and comes with its own command:
    - `cd <monarch-app>`
    - `poetry -C backend install -E api`
    - `monarch-api`
- add option to download specific version of solr kg
- bump minor version
- update testing a bit, disabled a couple until we can get a decent mock solr implementation 
    - they still run locally though, so before each commit/pr we should absolutely spin up a local solr and run `make test-backend` for now (not ideal, but. where we're at)


vincerubinetti:
This seems to affect mostly just the backend so my review here is perfunctory. Looks good.

Only thing I noticed is that you removed (all?) the install commands from the makefile, like before each lint/test/etc step. I assume you moved the install to somewhere else, otherwise you'd be getting errors when running these things, but I couldn't find where you moved them. Or maybe it's expected that anyone running things locally knows to install first, and also GitHub Actions still includes an install step.

glass-ships:
> This seems to affect mostly just the backend so my review here is perfunctory. Looks good.
> 
> Only thing I noticed is that you removed (all?) the install commands from the makefile, like before each lint/test/etc step. I assume you moved the install to somewhere else, otherwise you'd be getting errors when running these things, but I couldn't find where you moved them. Or maybe it's expected that anyone running things locally knows to install first, and also GitHub Actions still includes an install step.

yeah, this is a design choice i picked up from i think nico or harshad. basically to reduce repeated actions and output - we don't necessarily need to regenerate the model or reinstall the library every time we want to test the backend. 

so ideally the user would just `make install` once and be good

kevinschaper (approved):

PR #201, opened 2023-07-07
Mock solr

glass-ships:
Mock `SolrImplementation` using pytest/unittest to allow tests to run tests without a local running Solr instance. 

- Very basic at this point, does not mock `SolrService`, but probably should so we can actually test some of the logic of `SolrImplementation` and not purely getting pre-made fixtures


glass-ships:
Dropping this for now as mocking may not be the solution we're after.
Will make new PR to update testing in piece mail

PR #202, opened 2023-07-10
Add evidence count to schema

kevinschaper:
This adds evidence_count and and full_name fields to the schema. Our `latest` build is already using this schema, so this PR needs to go in so that we can update api-dev to the latest graph.


kevinschaper:
@vincerubinetti Can you help me out with the e2e tests? 

--edit - never mind, it was an easy fix once I tracked it down

glass-ships (approved):

PR #203, opened 2023-07-11
Redesign SolrImplementation, and update backend testing

glass-ships:
Methods in `SolrImplementation` should follow the design pattern:
```
from monarch_py.implementations.solr.solr_parsers import parse_thing
from monarch_py.implementations.solr.solr_query_utils import build_thing_query

def get_thing(args):                         
    query = build_thing_query(args)
    result = solr.query(query)           
    thing = parser(result)               
    return thing                         
```

Then independently we can test the parsing and query building methods. 
(we can also generate fixtures for the query, solr response, and datamodel for this)


kevinschaper (approved):
Thank you! this is a huge improvement

PR #238, opened 2023-07-20
Turn xrefs on node page into links

### LINKED ISSUE #191: External References links aren't going anywhere external
Right now all of the xrefs are going to back to self, like `https://monarch-app.monarchinitiative.org/node/GARD:0006322` and `https://monarch-app.monarchinitiative.org/node/Orphanet:98249`

Screenshot 2023-06-30 at 3 08 38 PM

We'll want to generate the urls from a prefix map, and I think we probably want to do it on the backend using https://github.com/cthoyt/curies

### LINKED ISSUE #237: Error loading node information when navigating via external links
Just noting an error when trying to navigate to a link to an external reference. Started by pulling up the page for Fanconi anemia (MONDO:0019391). Tried opening a new tab using the Orphanet:84 and OMIMPS:227650 links from the External References section and I get an "Error loading node information" error.

kevinschaper:
Uses curies library with the monarch context to supply an `external_links` collection with the node page response which includes an `id` and `url` for each xref. `url` is null when not available from the map. 

These are displayed on the node page using AppLink, same as before, except that the url is passed in, or swapped for an empty string when null.

Fixes #191 
Fixes #237


backend/src/monarch_py/datamodels/model.yaml
```diff
@@ -260,6 +266,10 @@ slots:
   evidence_count:
     description: count of supporting documents, evidence codes, and sources supplying evidence
     range: integer
+  external_links:
+    range: ExpandedCurie
```
glass-ships, line 271: maybe add a description clarifying what this is for /  how it differs from xrefs?

glass-ships (approved):

vincerubinetti:
Will take a look tomorrow

vincerubinetti (approved):

PR #276, opened 2023-08-14
Link to all association table subjects/objects if not a self-link

### LINKED ISSUE #267: subject and object entities (that aren't the page you're on) should be links
Right now, we aren't linking to any entities from the associations table:

Screenshot 2023-08-09 at 7 04 36 PM

In 3a, there were links to the other side of any association.
Screenshot 2023-08-09 at 7 05 14 PM

We should definitely at least link to the other side of associations, but I think it probably makes sense to change up our logic from 3a and just link to any entity that isn't the page we're already on. 

(thinking about how it also loses some of the sense of how each association relates to the entity page that we're on, it would be awesome if were able to make that association explicit, but I think that's a bigger change than we'd want to make right now)

kevinschaper:
Fixes #267

Adds association links for either side of an association as long as it wouldn't be a self-link

### DIFF
--- a/frontend/src/pages/node/AssociationsSummary.vue
+++ b/frontend/src/pages/node/AssociationsSummary.vue
@@ -28,7 +28,7 @@
             name: item.subject_label,
             category: item.subject_category,
           }"
-          :link="node.id === item.object"
+          :link="node.id !== item.subject"
         />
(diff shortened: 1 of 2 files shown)


vincerubinetti (approved):
As long as this behavior looks good to you in testing, looks good to me.

PR #285, opened 2023-08-22
change api root logic

### LINKED ISSUE #284: Add "relative" path to list of API urls?
Since hosting the UI alongside the rest of the v3 stack, I've noticed that all the options in the list below are absolute URLs.

https://github.com/monarch-initiative/monarch-app/blob/f4eb50c6220ee668cd5054f3df1f6e4609afe3e3/frontend/src/api/index.ts#L33-L38

Would it be possible to add an extra option, say "relative", that makes the API URL a relative URL, e.g. `{ relative: "/v3/api" }`?

With that in place, I'd be able to more easily test the UI container since it'll be running from the same host as the API.

vincerubinetti:
Closes #284 

- rework how api flags are handled. add "relative" mode for arbitrary domain.

### DIFF
--- a/frontend/README.md
+++ b/frontend/README.md
@@ -64,8 +64,9 @@ Which version of the Monarch API to utilize in the frontend, e.g.:
 URL param: `monarchinitiative.org/?api=dev`  
 Env var: `VITE_API=local`
 
-If not set, inferred from domain of URL.
-See `src/api` for the complete list of short API names, full API urls they map to, and related logic.
+If not set, inferred from URL.
+Flag gets directly inserted into the API URL like `api-beta.monarchinitiative.org`.
+See `src/api/index.ts` for related logic.
 
 **`mock`**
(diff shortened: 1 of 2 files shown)


glass-ships (commented):
yeah this looks good to me, it's a nice flexible option

glass-ships (approved):

falquaddoomi (approved):
Looks good, and it addresses my needs. Thanks!

kevinschaper (approved):

kevinschaper:
I want to move to relative urls `<host>.mi.org/api/v3` - but we shouldn't commit that change to the api until we have it working on the back end.

vincerubinetti:
Okay so I guess we'll merge for now and change it later. Also, I hope we decided on "next", but it's still "v3" at the moment, so I'll change the fallbacks to that.

PR #289, opened 2023-08-24
Update build-image.yaml, build docker image on release publish

kevinschaper:
Does this look right for also building a docker image when we publish a release?  (now that the version number isn't changed as a commit)


glass-ships (approved):
yeah that should work, good catch!

PR #291, opened 2023-08-25
Allow semsim compare of gene/disease by expanding to phenotype term sets

kevinschaper:
Initially, this PR is adding /entity/<ID>/compare/<other ID>, which is a nice explicit compare that anchors itself to an an entity, but we likely also need to support expanding gene and disease IDs in the regular compare endpoints


kevinschaper:
This won't end up being the approach that we take

PR #301, opened 2023-08-31
Add breadcrumb tracking for associations

### LINKED ISSUE #302: Show an explicit message rather than an empty pulldown when a node has no associations
Screenshot 2023-08-31 at 3 45 58 PM

MONDO:0004276 doesn't have any associations (at least, not that we show in the associations table), so it ends up with just an empty selector for choosing a type of associations to show. We could show some kind of explicit message to say that it doesn't have associations.

vincerubinetti:
- add back in breadcrumb tracking for associations, including adding "implicit" association/link when clicking on a subject/object that is not the current node but instead a sub-class of it.
- to breadcrumb section, add tooltips for "steps backward in history", and fade "implicit" breadcrumbs.
- fix some css class name collisions from a previous PR
- fix blank () taxon label in node badge component
- fix search filtering. dont pass param to api at all if all/none selected. re-introduce showCounts prop in multi select component to allow global showCounts instead of per-instance showCounts.
- add explicit blank messages for histopheno and association counts sections

### DIFF
--- a/frontend/src/components/AppDetail.vue
+++ b/frontend/src/components/AppDetail.vue
@@ -8,7 +8,7 @@
     align-h="left"
     direction="col"
     gap="small"
-    :class="['detail', { big }]"
+    :class="['detail', { full }]"
   >
     <AppFlex gap="small" align-h="left">
       <AppIcon v-if="icon" :icon="icon" class="icon" />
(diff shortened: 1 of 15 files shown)


glass-ships (approved):
nice!

vincerubinetti:
@kevinschaper nudge. The breadcrumbs might require more careful thought, but I also added some important bug search fixes that I'd like to merge. Maybe I can just hide the breadcrumbs section for now and we can uncomment it later?

kevinschaper (approved):

Issue #270, opened 2023-08-10
multi-entity endpoint


oneilsh:
I'm looking for a blunt-instrument API endpoint to support the R package. What I'd like to do is provide a list of entities, and a list of entity categories that they might connect to, and get those associations back in bulk.

Here's an example request:

```yaml
entities:
  - MONDO:103135
  - HGNC:1356
categories:
  - biolink:Gene
  - biolink:PhenotypicQuality
limit_per_group: 10
offset_per_group: 0
``` 

And an example result:

```yaml
entities:
    - id: "MONDO:103135":
      name: "Some Rare Syndrome"
      category: "biolink:Disease"
      associated_entity_categories:
            - category: "biolink:Gene":
              total: 140
              associations:
                   - id: "HGNC:13513"
                     name: "HJPR1"
                     predicate: "gene_correlated_with"
                     association_category: "biolink:CorrelatedGeneToDiseaseAssociation"
                     publications: ...
                   - id: "HGNC:161615"
                     name: "LN3"
                     predicate: "gene_causes"
                     publications: ...
                     association_category: "biolink:CausalGeneToDiseaseAssociation":
            - category: "biolink:PhenotypicQuality":
              total: 30
              associations:
                   - id: "HP:13951134"
                     name: "Short stature"
                     predicate: "phenotype_associated_with"
                     publications: ...
                     association_category: "biolink:DiseaseToPhenotypicFeatureAssociation"
                   - ...
                   - ...
    - id: "HGNC:1356":
      category: "biolink:Gene"
      connected_entity_categories:
           - category: "biolink:Gene":
             total: 8
             associations:
                  - id: "HGNC:1352"
                    name: "JJNS2"
                    predicate: "interacts_with"
                    publications: ...
                    association_category: "biolink:GeneGeneInteractionAssociation"
                   - ...
                   - ...
           - category: "biolink:PhenotypicQuality":
             total: 5
             associations:
             ...
```

The limit and offset would be applied per combination, so the maximum number of results should be `#entities * #categories * limit_per_group`. If necessary the total number of entities, categories, and limit can be limited to keep performance. As the offset increases it's expected that some groups will run out of results before others.

kevinschaper:
I had been thinking about this in terms of using solr's result grouping, and churning a little on how I could hack multi-level grouping, when solr only supports grouping by a single field... but maybe it makes more sense to just build this response by doing a separate solr query for each of the permutations of entities and categories specified. 

[...]

PR #310, opened 2023-09-07
second pass at multi-entity assocations endpoint

glass-ships:
Addresses #270 

no CLI implementation yet, still working out the under-the-hood stuff.

currently just manually testing/checking output with:
```python
from monarch_py.implementations.solr.solr_implementation import SolrImplementation
from monarch_py.utils.utils import to_json
from pprint import pprint

si = SolrImplementation()

results = si.get_multi_entity_associations(
    entity=["MONDO:0012933", "MONDO:0005439"], counterpart_category=["biolink:PhenotypicQuality", "biolink:Disease"]
)
print(len(results))
count = 0
for i in results:
    count += 1
    to_json(i, file=f"test_multi_endpoint_{count}.json")
```


glass-ships:
is this maybe closer to what you were imagining @oneilsh ?

oneilsh:
@glass-ships yes I think so!

oneilsh:
@glass-ships I was just noticing the if `entity == None: continue` lines; I think if there's not a good match for a requested entity then something should be returned in that vs an entity that doesn't have any associations. Currently it looks like a lack of an entity in the response means it couldn't be found, while an entity without associations would get an entry in the return that is empty right? I wonder if each entity shouldn't have a basic "status" field to indicate if it wasn't found, or maybe the category for the entity should be null or something. I dunno, open to suggestions.

```
entities:
    - id: "MONDO:NOSUCHID":
      name: null
      category: null # like this maybe?
      status: "not found" # or like this?
    - id: "MONDO:113513"
      ...
```

glass-ships:
hey @oneilsh, sorry again for the delay between check-ins. 
how does the most recent commit look? 
I added a bit before the continue statement that may address this concern, for example, running:
```python
results = si.get_multi_entity_associations(
    entity=["MONDO:0012933", "MONDO:0005439", "MANGO:0023456"], 
    counterpart_category=["biolink:PhenotypicQuality", "biolink:Disease"]
)
```

would return this for the last entity ID:
```python
{
    "limit": 20,
    "offset": 0,
    "total": 0,
    "id": "MANGO:0023456",
    "name": "Entity not found",
    "associated_categories": []
}
```

oneilsh:
@glass-ships I like it, thank you! I'm excited, this is going to be a very friendly endpoint (both both humans and AI models :P). In the future we may want to consider different options for sorting results (e.g. by evidence count, or betweenness centrality, or embedding similarity to a given query string, ...), but that can be for another day.

kevinschaper (approved):
This is awesome, thank you @glass-ships!

glass-ships:
should i add a CLI command before merging this? @kevinschaper / @oneilsh

kevinschaper:
Nah, I think we should just merge it in, but maybe also make an issue to add the cli endpoint.

PR #313, opened 2023-09-12
fix feedback form, improve request error logging

vincerubinetti:
- add appropriate headers to feedback form to fix error
- remove error response type from feedback endpoint because request should throw error anyway and we'll never reach `if ("error" in data)`
- rearrange request wrapper func such that response is logged even if response not OK or error parsing
- misc formatting, and throw new Error -> throw Error

### DIFF
--- a/frontend/src/api/index.ts
+++ b/frontend/src/api/index.ts
@@ -132,15 +132,22 @@ export const request = async <Response>(
     /* make new request */
     response = await fetch(url, options);
 
-  /** check response code */
-  if (!response.ok) throw new Error(`Response not OK`);
+  /** capture error for throwing later */
+  let error = "";
 
-  /** add response to cache */
-  if (request.method === "GET") cache.set(id, response.clone());
+  /** check response code */
+  if (!response.ok) error = `Response not OK`;
 
   /** parse response */
-  const parsed: Response =
-    parse === "text" ? await response.text() : await response.json();
+  let parsed: Response | undefined;
+  try {
+    parsed =
+      parse === "text"
+        ? await response.clone().text()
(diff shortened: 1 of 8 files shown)


glass-ships (approved):
nice, thanks for handling that so quick!
(out of curiosity, why `throw new Error` -> `throw Error`?)

frontend/src/api/index.ts
```diff
@@ -132,24 +132,37 @@ export const request = async <Response>(
...
-      response,
+      response: response.clone(),
     });
 
+  /** throw error after details have been logged */
+  if (error || parsed === undefined) throw Error(error);
+
+  /** add response to cache */
+  if (request.method === "GET") cache.set(id, response.clone());
+
```
falquaddoomi, line 166: I see that you have a lot of calls to `response.clone()` in this code. Out of curiosity, why not clone it once and use that single cloned version for all these references?

vincerubinetti:
> nice, thanks for handling that so quick! (out of curiosity, why `throw new Error` -> `throw Error`?)

They're equivalent and the latter is just more terse.

Let's merge this even though tests are failing. They are passing locally for some reason, so there's something brittle there. I want to merge because the feedback form working is very important right now.

falquaddoomi (approved):
I understand it's important to get this in and I don't see anything wrong with the code, so I'll approve for now.

frontend/src/api/index.ts
vincerubinetti: This was left over from some re-arranging I was doing in a hurry. Only the `await response.clone().text()` clones are needed since the body is being consumed. Fixed in #315

PR #315, opened 2023-09-12
Fix unit test failures on gh-actions

vincerubinetti:
- fixes regression in `apiCall` util call due to unknown cause. add arbitrary timeout (see issue linked in code comment) in lieu of adding whole new package `@testing-library/vue` just for one function `waitFor` and rewriting all unit tests that use `apiCall`.
- update all packages to latest versions


frontend/unit/setup.ts
```diff
@@ -38,6 +38,7 @@ export const apiCall = async (): Promise<void> => {
    */
   await sleep();
   await sleep();
+  await sleep(10);
```
falquaddoomi, line 42: Perhaps a comment here about why it's 10 seconds, in addition to the two non-parameterized calls above, would be nice.

falquaddoomi (approved):
Looks good; I left a note about clarifying a bit of code, but it all looks reasonable. Seems the tests are passing now, too, so kudos. Nice catch on the configuration key for "prettier-plugin-css-order", too.

frontend/unit/setup.ts
vincerubinetti: Unfortunately, the 10ms is arbitrary and I really don't know why the error is happening. I can't replicate it locally.

I'll add a generic comment to this effect, also leave this PR open a little longer (maybe til next week) to keep investigating.

vincerubinetti:
Merging this so PRs stop failing, but I'm setting a reminder for myself to keep investigating.

PR #317, opened 2023-09-15
Add semantic search

kevinschaper:
Support for loading phenio.db from a file path (with a download as a fallback), along with an endpoint to call oaklib's semantic search

- [x] Change up to using pystow to fetch phenio db, pointing at the monarch ingest version with term_associations
- [x] Support for getting phenio.db location from the environment

- [ ] Add API endpoint


kevinschaper:
This magic is happening in #499 now

PR #333, opened 2023-09-20
tweak unit test setup funcs

vincerubinetti:
Addresses #315 

I've done extensive testing with [tmate on github actions](https://github.com/mxschmitt/action-tmate), messing with sleep locations/counts, package versions, test setup/teardown, logging at various places, etc. I also managed to reproduce the bug locally, seemingly by updating all packages, clearing yarn cache, and reinstalling (even though I'm fairly sure I did this before).

I'm more certain now that the error is due to `msw` arbitrarily wrapping things in promises. Now the magic number of sleeps seems to be 3, and may need to be increased in the future depending on `msw` implementation details. I've added a comment about this. I've also determined I could get rid of some test setup/teardown functions.

### DIFF
--- a/frontend/unit/setup.ts
+++ b/frontend/unit/setup.ts
@@ -2,27 +2,14 @@ import { nextTick } from "vue";
 import type { Component } from "vue";
 import { cloneDeep } from "lodash";
 import { setupServer } from "msw/node";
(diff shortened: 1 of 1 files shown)


glass-ships (approved):
definitely some weird behaviour, nice find!

Issue #325, opened 2023-09-19
Automate update of GCP Docker Swarm


glass-ships:
It'd be pretty handy if we could use GHActions to automate the process of updating the docker images used in the various GCP VMs that serve the UI and backend (Nginx, api) 

This may look something like:
```bash
$ gcssh monarch-v3-api && docker pull us-central1-docker.pkg.dev/monarch-initiative/monarch-api/monarch-api:latest
# and separately
$ gcssh monarch-v3-manager && docker swarm update
```

glass-ships:
Being worked on in #339 
Related issues: 
- monarch-initiative/monarch-stack-v3/issues/18
- monarch-initiative/monarch-stack-v3/issues/19

glass-ships:
Seems to be working via GH Actions despite ongoing issues manually provisioning from within MI GCP. 
Create new issue if a new problem arises

PR #339, opened 2023-09-21
Automate updating GCP Docker swarm

glass-ships:
Addresses #325


glass-ships:
Attempted to merge workflows to build/push frontend and backend images, 
then run separate work flow to update gcp services. 

but something seems to be wrong with the `on:` syntax in the build-images workflow. 
will investigate further later

glass-ships:
Appears to be a potential problem with the docker service update command that requires a full reset via terraform/ansible:
```
1/1: no suitable node (scheduling constraints not satisfied on 2 nodes; 1 node …
overall progress: 0 out of 1 tasks
service update paused: update paused due to failure or early termination of task qtr8cqtrn88fnbshv0aobrjxd
Error: Process completed with exit code 1.
```

kevinschaper:
We probably need more memory than 14gb (maybe 16?) to get this to fully succeed, but I don't think there's anything we need to change about this PR, so we should bring it in

kevinschaper (approved):

PR #345, opened 2023-09-25
table format not implemented for Node in CLI

### LINKED ISSUE #215: Table format breaks on entity with --extra
monarch-py doesn't know how to flatten a nested object into more columns for a table

glass-ships:
- Closes #215 

We can look at this in more detail later, but for now we'll just not implement it, as the extra columns aren't particularly useful in a table format


kevinschaper (approved):
The Fastapi update looks great too!

PR #353, opened 2023-09-29
Fix api dying

glass-ships:
Adds a startup script to use in dockerfile instead of a list style CMD

### DIFF
--- a/backend/start_api.sh
+++ b/backend/start_api.sh
@@ -1,8 +1,8 @@
 #!/usr/bin/bash 
 
-# check for uvicorn workers env var, default to 8
+# check for uvicorn workers env var, default to 1
 if [ -z "${UVICORN_WORKERS}" ]; then
-    UVICORN_WORKERS=8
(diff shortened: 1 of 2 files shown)


backend/start_api.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/bash 
+
+# check for uvicorn workers env var, default to 1
+if [ -z "${UVICORN_WORKERS}" ]; then
+    UVICORN_WORKERS=1
+fi
```
falquaddoomi: This works as-is, but there's a shorthand version in bash for setting a variable to a default if it doesn't exist:

```suggestion
UVICORN_WORKERS=${UVICORN_WORKERS:-1}
```

The suggestion attempts to use `${UVICORN_WORKERS}` from the environment first, then if it's empty or unset uses the default value `1`.

falquaddoomi (approved):
Looks good! I made a small suggestion, but it's not critical for approval.

PR #356, opened 2023-09-29
Load phenio.db from /data if available, use pystow to fetch from monarch data bucket if not

kevinschaper:
This should stop our multiple pystow download problem (and mean no download on startup at all, happily)


falquaddoomi (approved):
Nice! I haven't tried it, but I read through the code and it looks reasonable to me.

glass-ships (approved):

PR #360, opened 2023-10-02
Basic phenogrid replacement

vincerubinetti:
Will summarize what this PR does tomorrow, but you can start taking a look at it now. See `/phenogrid` and `/testbed` and the readme section added.

- add debug log step to every test, so re-running a failed job with "debug logging enabled" in github actions, it will pause and allow you to ssh into the runner with tmate
- added readme instructions for using phenogrid widget
- in playwright, pass console logs to cli logs
- add more msw api mocking unhandled request exceptions. convert them to regexes for flexibility
- update pheno explorer compare fixture data
- fix bug where tab title momentarily says "undefined" on first page load
- remove old phenogrid css, libraries, and component
- derive state needed for phenogrid from new monarch compare endpoint
- make new basic phenogrid component that basically just displays a table/grid of colored cells with tooltip information
- add the phenogrid iframe widget to testbed to mimic how it would behave if embedded in another site
- add /phenogrid route with just the phenogrid component on the page (with no header/footer/etc) that can be used as an embeddable iframe
- make all routes lazy loaded (on demand). this way people using the /phenogrid embed don't have to request the entire app's code, but instead (hopefully) only the route and code needed for the widget


kevinschaper:
It's looking pretty strange on safari for me (but looks great on chrome)

Screenshot 2023-10-03 at 8 40 25 AM

vincerubinetti:
Safari strikes again. Looks like a long standing (since 2012) bug that [it doesn't support writing-mode in table cells](https://bugs.webkit.org/show_bug.cgi?id=65917). Unbelievable. Working on fixing.

kevinschaper:
I very much appreciated the 10 yr old comments pointing out that even IE supports this.

kevinschaper (approved):

Issue #380, opened 2023-10-10
Change logging mode for dev & production servers


kevinschaper:
Right now we're getting a lot of log noise when solr_service prints out requests going out to solr. I was tempted to remove that, but  obviously it makes more sense to not run our API in debug mode all of the time.

glass-ships:
see #381

glass-ships:
https://pawamoy.github.io/posts/unify-logging-for-a-gunicorn-uvicorn-app/

This may be worth a read

kevinschaper:
Moving to 2023-11 release, though we may end up bumping it further back

kevinschaper:
Closing, it would be nice to sort out our python logging at some point, but right now I don't think we'd get enough benefit from it to dig in on that.

PR #381, opened 2023-10-10
set default log level to info

glass-ships:
Addresses #380 

Sets default log level to INFO

Can be set via `monarch_py.api.config`


backend/src/monarch_py/utils/utils.py
```diff
@@ -58,10 +58,9 @@ def dict_diff(d1, d2):
 
 def set_log_level(log_level: str):
     """Sets the log level for the application."""
-    import loguru
+    from loguru import logger
 
-    loguru.logger.remove()
-    loguru.logger.add(sys.stderr, level=log_level)
+    logger.configure(handlers=[{"sink": sys.stdout, "level": "ERROR"}])
```
kevinschaper: It looks like this ends ignoring the log level that's passed in?

glass-ships: it totally was yeah, my bad

glass-ships:
I can't really remember exactly where we left off on this, maybe @falquaddoomi or @kevinschaper can recall the details. 

IIRC, this sort of worked locally, but the effects didn't make their way to the GCP logs, which kinda defeated the main purpose 

But TLDR, I think we'll retire this attempt and circle back at a later date with a different approach

Issue #358, opened 2023-10-02
Documenting Monarch Data Archiving and PURL strategy 


sagehrke:
From Monarch Data call on 2023-Sept-28, the following was identified as our **Monarch Data Archiving and PURL strategy** moving forward:

1. [data.mi.org](http://data.mi.org/) is being equipped with a fairly stable file structure (great), but: better safe then sorry and use .htaccess redirects to complement this for stability. Kevin comments also that this is a great way to document what kind of files are important to the outside world.
2. No one argued against [data.mi.org](http://data.mi.org/) stability. It seems there is a positive feeling that it can be maintained on a reasonable time horizon (5-10 years). No one seems to be against managing our PURls on [data.mi.org](http://data.mi.org/).
3. Archiving is a very different issue; while zenodo isnt great for access to individual files, it is a great place for “permanence”.
4. TDLR: we will manage monarch purls on [data.mi.org](http://data.mi.org/) with an .htaccess file, and when the project dies, we redirect the PURLs to the Zenodo mega dumps.

- **When this strategy is finalized, please add it to the Monarch Technical Documentation here: https://monarch-initiative.github.io/monarch-technical-documentation/**

sagehrke:
@kltm responded to above:
- I think I have a slightly different read on 2 and 4, although it may be splitting hairs, so please take with a grain of salt.
- For 2, I'm not sure that PURLs are the best way of framing data stability. My understanding was that if stability can be maintained, a "formal" PURL system may not be strictly necessary as "stable" URLs are good enough. In my mind, a PURL is likely run by a party with some kind of separate commitment to exist longer than any individual that participates in it. That could be the BBC making sure that URLs for old programs are still usable or that OBO makes usable archive links for ontologies that become unsupported. Or, another way: grant funding is stable, but not permanent--time horizons are important and 5 years (~a grant cycle) is very different than 10.
- For 4, depending on the granularity, I don't think it is really possible. Zenodo works best on project-level concepts, so a PURL to an ontology file would not really work redirected to Zenodo except as a "good luck with that". It's mixing archive with PURLs and end-user browsability.

@matentzn responded: 
[...]

PR #424, opened 2023-10-20
Adds a redirect to handle external links to previous node pages

### LINKED ISSUE #417: Add support for /disease/<ID>, /gene/<ID> etc
I'm seeing lots of requests coming in for `/disease/MONDO:0012790` from legit sources like `"https://0-www-ncbi-nlm-nih-gov.brum.beds.ac.uk/medgen/96594` and that url only goes to the home page right now, we can definitely get them to the right place.

kevinschaper:
Should fix #417

### DIFF
--- a/services/nginx/config/default.conf
+++ b/services/nginx/config/default.conf
@@ -1,5 +1,5 @@
 upstream api {
-    server localhost:8000;
+    server api:8000;
 }
 server {
     listen 80 default_server;
--- a/services/nginx/config/default.conf
+++ b/services/nginx/config/default.conf
@@ -1,5 +1,5 @@
 upstream api {
-    server api:8000;
+    server localhost:8000;
 }
 server {
     listen 80 default_server;
@@ -26,7 +26,7 @@ server {
     }
 
     # redirect /<category>/<curie> to /<curie>
-    location ~^/[a-z]+/([^:]:[^:])$ {
+    location ~^/(gene|disease|phenotype|anatomy|pathway)/([^/:]:[^/:])$ {
         absolute_redirect off;
         return 301 /$2;
     }
--- a/services/nginx/config/default.conf
+++ b/services/nginx/config/default.conf
@@ -26,7 +26,7 @@ server {
     }
 
     # redirect /<category>/<curie> to /<curie>
-    location ~^/(gene|disease|phenotype|anatomy|pathway)/([^/:]:[^/:])$ {
+    location ~ ^/(gene|disease|phenotype|anatomy|pathway)/([^/:]:[^/:])$ {
         absolute_redirect off;
         return 301 /$2;
     }
--- a/services/nginx/config/default.conf
+++ b/services/nginx/config/default.conf
@@ -27,7 +27,8 @@ server {
 
     # redirect /<category>/<curie> to /<curie>
     location ~^/[a-z]+/([^:]:[^:])$ {
-        return 302 /$2;
+        absolute_redirect off;
+        return 301 /$2;
     }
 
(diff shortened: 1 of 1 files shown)


falquaddoomi:
If this is intended to redirect to a page on the frontend, IMHO the frontend router should handle this, not NGINX. All requests that aren't for the API already go to the frontend, with the intent that the frontend router would parse the request path and display the correct page. @vincerubinetti, any opinions there?

services/nginx/config/default.conf
```diff
@@ -25,6 +25,11 @@ server {
         return 302 /v3$request_uri;
     }
 
+    # redirect /<category>/<curie> to /<curie>
+    location ~^/[a-z]+/([^:]:[^:])$ {
+        return 302 /$2;
```
falquaddoomi: This should probably also include the `absolute_redirect off;` directive like the "/api" block above, otherwise NGINX will attempt to construct an absolute URL and likely fail to produce a working one in non-production environments.

kevinschaper:
I initially started in the router, and then talked myself into the idea that since the app doesn't produce these urls, it shouldn't necessarily have the job of handling them - but I can just as easily talk myself out of that idea. 

I suppose I'm also getting a start at using nginx redirects in preparation for #358, but that doesn't really related to this other than the chance for me to poke at redirects.

falquaddoomi:
> I initially started in the router, and then talked myself into the idea that since the app doesn't produce these urls, it shouldn't necessarily have the job of handling them.

Fair point; I suppose for me it makes sense for the frontend to handle it, since it's a URL intended to (eventually) show a page in the frontend, but like you, I could argue it either way.

vincerubinetti:
I don't mind either way.

services/nginx/config/default.conf
```diff
@@ -25,6 +25,12 @@ server {
         return 302 /v3$request_uri;
     }
 
+    # redirect /<category>/<curie> to /<curie>
+    location ~^/[a-z]+/([^:]:[^:])$ {
```
falquaddoomi: (Thanks for the quick fix on my last comment, and sorry for bringing up the same line so soon after!)

I was thinking through this regex a bit and I'm a little concerned that it may be too permissive; it'd catch, e.g., `/<alphabet-letters>/<anything with a single colon in it, even multiple levels down>`. I don't know how many possible values of `<category>` there are, but perhaps if there aren't too many we could explicitly name them here?

Also, I assume you want to match more than a single non-colon on either side of the colon, so you probably would want to add a `+` to each. Finally, I'm not 100% sure, but I think you might need a space between the tilde and the initial caret. Here's how I'd modify it:

```suggestion
    location ~ ^/[a-z]+/([^/:]+:[^/:]+)$ {
```

This also prevents it from descending into multiple levels, but you can remove the slash from the first character group to get rid of that behavior. I'd prefer it even more if the categories were explicitly specified, and the patterns on either side of the colon could be limited to just the legal characters for those identifiers.

kevinschaper: I was having the same thoughts about the super permissive catch for categories, I could enumerate the ones that would actually work well, and if somebody comes in with something like `/case/MONARCH:123` it might as well go to search anyway. 

I think I'll also move the redirect over to the router, it's so much easier to test that way.

kevinschaper: I take it back, I think I'm sticking with the nginx redirect, I struggled for a bit with the Vue router syntax and decided that I wanted to finish it more than I wanted to track down how to do a 301 redirect from Vue.

falquaddoomi (approved):

PR #461, opened 2023-11-02
Schema support for SSSOM Mappings and association grouping_key column

kevinschaper:
This one is going to be a little rocky to merge and deploy, since we'll also need to load new data to dev.mi.org at the same time so that the schema will match


backend/src/monarch_py/datamodels/model.yaml
```diff
@@ -193,6 +199,16 @@ classes:
     is_a: FacetValue
     slots:
       - id
+  Mapping:
+    description: >-
+      A minimal class to hold a SSSOM mapping
+    slots:
+      - subject_id
```
glass-ships, line 207: i still have the same question - aren't `subject` and `subject_id` redundant?

glass-ships: same thing with `predicate_id` and `object_id`

glass-ships (approved):

backend/src/monarch_py/datamodels/model.yaml
kevinschaper: I definitely agree that it's a little weird and unfortunate, but this little corner of our schema needs to match SSSOM's schema, which doesn't quite line up with the biolink model.

Issue #348, opened 2023-09-28
Bang our own drum 


pnrobinson:
It would be great to have the front page of the Monarch App show a list of "our" ontologies with links to the specific resources -- Mondo, Hpo, Uberon, Oba, etc

sagehrke:
I agree with this! Making the Monarch app the centralized location to find all of the many monarch tools, ontologies, etc would be great. **Can we make a concrete list of what to include and the best way to integrate it into the UI?** I feel like it would be great to have a new dropdown that is "Products" or something like that, and we can list everything with a short description and link to the content. The homepage could have a few of the Flagship Monarch projects (based on input from the Monarch leadership) listed with quick button links. 

Off the top of my head, it would be nice to have links to: HPO, Phenopackets, Mondo, Exomiser, LinkML, uPheno, OAK, Boomer, SSSOM, POET, PhEval, Uberon, OBO Academy training, KG-Hub(?), OBA

sagehrke:
@glass-ships Please create the following at monarchinitiative.org: 
- Add another button to the top ribbon that is titled "Resources"
- On this new “Resources” landing page, you can follow the same formatting as the "About" page with the circle icon links. 
- Organize each resource into one of three categories: Standards, Ontologies, Tools
- Under each category, list the resource (from the Flagship product list I have shared with you). Each resource will have a button to an outside site that will either be a generic icon or the official project logo (preference is to use a project logo if it exists). Note: Logos can be found here: https://github.com/jmcmurry/closed-illustrations/tree/master/logos
- Each Resource will have the resource’s name, the link to the external site for the resource, and then a one-liner tagline for what the resource is.

glass-ships:
Added some basic taglines for each resource - @sagehrke / @monicacecilia feel free to review and revise as appropriate
Also opened an issue for a textless mondo icon, and marked a few that have no logo (mostly for my own reference)

vincerubinetti:
Does this need to be in the header? Can it just be a featured link at the top of the About page, or on the home page? The header is precious real estate.

sagehrke:
Hi @vincerubinetti - yes, please. We'd like the "Resources" to be in the header, not nested on another page.

glass-ships:
done in #463 - preview available [here](https://deploy-preview-463--monarch-app.netlify.app/resources)
please review at your convenience

PR #463, opened 2023-11-02
Add Resources (Bang Our Own Drum)

glass-ships:
Addresses #348 

- [x] Adds resources page and endpoint to router
- [x] Adds some icons
- [x] Need to figure out icons for ones that don't have one (maybe @vincerubinetti will have ideas) ((Mondo should get a textless version eventually, i just don't have illustrator))


vincerubinetti:
There's color contrast problems here. Try adding these styles to the app icon component, under `.custom`:

```css
filter: brightness(0);
```

frontend/src/pages/resources/PageResources.vue
```diff
@@ -0,0 +1,59 @@
+<template>
+  <AppSection>
+    <AppHeading>Resources</AppHeading>
+    <p>
+      This page contains links to resources and tools that are central to the
```
vincerubinetti: ~This page contains links to~ Resource and tools ~that are~ central to ~the~ Monarch Initiative.

vincerubinetti:
Otherwise this looks good and is how the other pages are done. I don't see anything that would be causing some icons to inconsistently load. It sounds like it might just be a bug with the vite hot-module reloading and the vite svg loader plugin. I don't see the behavior on the netlify preview.

glass-ships:
> Otherwise this looks good and is how the other pages are done. I don't see anything that would be causing some icons to inconsistently load. It sounds like it might just be a bug with the vite hot-module reloading and the vite svg loader plugin. I don't see the behavior on the netlify preview.

that might be true, i only noticed it locally anyway. thanks vince! i'll make those changes here shortly, then add the mondo icon when it comes in. any ideas about the ones that don't have logos? (like mapping commons, ontoGPT, etc)

vincerubinetti:
If you can't find logos anywhere on the internet, maybe just leave them as initials? (Just use a kebab-case icon name, without the file existing.)

vincerubinetti:
Screenshot 2023-11-03 at 1 05 08 PM

You'll also need to change the header breakpoint so this does not happen.

glass-ships:
> Screenshot 2023-11-03 at 1 05 08 PM
> You'll also need to change the header breakpoint so this does not happen.

any idea what i'm looking for in particular? sorry, i don't immediately follow

vincerubinetti:
`$wrap` in the header component. It is the screen width below which the header collapses into a smaller layout.

sagehrke:
Thank you for this draft @glass-ships! A few updates to consider:

- Can @monicacecilia write a blurb for the top explaining what Resources means a bit more than the one-liner currently on the preview? I am thinking something to the effect of "Monarch is more than the KG you see here...it is a whole compendium of resources for etc etc etc...find links to our flagship products app or GitHub repos below" Perhaps ending this paragraph with a button link to our awesome new detailed documentation that @matentzn is leading the charge on: https://monarch-initiative.github.io/monarch-documentation/
- I suggest we **remove** "Monarch KG & API" and "Monarch App" buttons on this page. While they are flagship, the info these may provide could instead be embedded into the /About page since that is all about the Monarch App, if it isn't there already (we can make another ticket for this if it is agreed that they should move - don't worry about doing it here). 
- I am still contemplating if I like the look of the buttons that we don't have logos for. I'd like to hear other people's opinions.  
- Update the "Standard" title to be plural "Standards"

glass-ships:
> I suggest we remove "Monarch KG & API" and "Monarch App" buttons on this page. While they are flagship, the info these may provide could instead be embedded into the /About page since that is all about the Monarch App, if it isn't there already (we can make another ticket for this if it is agreed that they should move - don't worry about doing it here).

there isn't really information added by either of these buttons. the only thing we don't do in the /about section is link to the `monarch-documentation` site itself, which we could do. but the Sources button links to the ingest docs, and the API button links to the same docs as the resources/Monarch KG & API button.

tldr; i support removing them, and maybe adding somewhere a link to the pages formerly known as technical docs

monicacecilia (approved):
ack! missed this one. 🙈

monicacecilia:
Dear @glass-ships  👀 👇 

> * Can @monicacecilia write a blurb for the top explaining what Resources means a bit more than the one-liner currently on the preview? I am thinking something to the effect of "Monarch is more than the KG you see here...it is a whole compendium of resources for etc etc etc...find links to our flagship products app or GitHub repos below" Perhaps ending this paragraph with a button link to our awesome new detailed documentation that @matentzn is leading the charge on: https://monarch-initiative.github.io/monarch-documentation/

Regarding a blurb at the top, we can update the content later - it is not a blocker. 

For later, we can add: 

_The Monarch Initiative is an international consortium that leads key global standards and semantic data integration technologies. Monarch resources and integrated data are also foundational to many downstream applications and contexts; we work closely with a variety of stakeholders and resource-development communities to capture feedback and make improvements._
_To maximize utility and impact, the Monarch platform is composed of multiple open-source, open-access components. We promote provenance and transparency, enhanced use of standards and new technologies and improved data accessibility, end-user utility, and data submission. Below, we highlight a collection of our flagship projects. Learn more about the complete suite of Monarch resources on our organization's [documentation pages](https://monarch-initiative.github.io/monarch-documentation/)._




> * I suggest we **remove** "Monarch KG & API" and "Monarch App" buttons on this page. While they are flagship, the info these may provide could instead be embedded into the /About page since that is all about the Monarch App, if it isn't there already (we can make another ticket for this if it is agreed that they should move - don't worry about doing it here).

Thank you. Let's keep them. It is not redundant, and if someone only visits this page, they will have a comprehensive overview. Thanks!



> * I am still contemplating if I like the look of the buttons that we don't have logos for. I'd like to hear other people's opinions.

They look okay. Let's keep them as is. We can ask Julie for more logos. Not a blocker.

glass-ships:
I see an approval, but because we moved to pulling the contents for this page from [monarch docs](https://github.com/monarch-initiative/monarch-documentation/blob/main/src/docs/resources/monarch-app-resources.json), we now need to make sure the content we need here exists in that repository. 

What has been requested is: 
```
Standards = [
    "Phenopackets",
    "SSSOM",
    "Mapping Commons"
]

Ontologies = [
    "HPO",
    "MONDO",
    "uPheno"
]

Tools = [
    "OAK",
    "OntoGPT",
    "CurateGPT",
    "LinkML",
    "Semsimian",
    "Exomiser",
    "LIRICAL"
]
```

While what currently exists in that repository does not match.

We'll need to make sure monarch docs is updated accordingly before we bring this PR in. Perhaps @matentzn can help with that process as he's taken on the task of implementing the generation of that document thus far

matentzn:
Don't worry, this list will change 5 times in the next months. SemSimian and OAK are missing for now but we can add them after the vote (more likely will be that they are not added, or that if they are added, 6 other things will be added alongside, like ODK).

PR #468, opened 2023-11-03
Convert new phenogrid to svg

vincerubinetti:
- update phenogrid readme instructions
- remove some unused fixture json files
- manually add really large phenogrid compare fixture json
- upgrade all packages to latest version. msw 1.0 to 2.0 (only major bump). refactor to accommodate.
- add some missing toLocaleString() to numbers
- change phenogrid component from table to svg. pros and cons discussed [here](https://github.com/monarch-initiative/monarch-app/pull/372)
- touchup behavior and look of phenogrid. add basic info tooltip. add transpose feature. rework size event emitting for widget use. download as svg instead of png
- touchup node page associations section
- fix small bug where clicking association count link in overview section does not scroll to associations section
- tweak how async route loading is done to avoid flash of homepage while loading
- add and tweak util funcs

Please take a look at the fixture data is being generated for the frontend. If it's not being imported at the top of `/fixtures/index.ts`, it's not being used and doesn't need to be generated.


vincerubinetti:
Let add/fix a few more things before merging, and I will re-ping @kevinschaper when ready.

vincerubinetti:
@kevinschaper ready

kevinschaper (approved):

PR #474, opened 2023-11-15
Add deprecated slot, down rank deprecated search results

kevinschaper:
Adds deprecated so that we can handle obsolete ontology terms properly


glass-ships (approved):

kevinschaper:
I added #478 to address the ugly `iri` field stripping that I'm adding here

PR #486, opened 2023-11-23
Simplify loading of converter in CurieService

cthoyt:
Uses newer code from `prefixmaps` that directly instantiates a converter


cthoyt:
@glass-ships can you approve this for CI?

glass-ships:
(we can ignore the unit test failing, it's unrelated and i believe vince is already looking into it)

glass-ships (approved):

PR #494, opened 2023-12-04
Replace CuriesService with direct usage of `curies.Converter`

cthoyt:
This PR follows up on #486 by fully replacing the CuriesService class and all functionality that makes thin wrappers around `curies.Converter` with direct usage.


glass-ships:
apologies this seems to have gotten lost in the weeds! 

@cthoyt would you be willing to take a look at the merge conflict? 

also just wondering - does the issue we ran into with 0.2.0 apply here, or should this work with 0.1.7 and/or will it require any changes when we update after resolving that?

cthoyt:
Merge fixed. We shouldn't be running into any version related issues due to this change

glass-ships (approved):

PR #495, opened 2023-12-06
Updating to Pydantic v2

### LINKED ISSUE #260: Updating to Pydantic V2
Koza will need this looked at as well, along with the following:
```
monarch-app/backend (similarity)
><glass@rocinante> poetry show pydantic
 name         : pydantic                                                        
 version      : 1.10.12                                                         
 description  : Data validation and settings management using python type hints 

dependencies
 - typing-extensions >=4.2.0

required by
 - bioregistry <2.0
 - curies <2.0
 - fastapi >=1.6.2,<1.7 || >1.7,<1.7.1 || >1.7.1,<1.7.2 || >1.7.2,<1.7.3 || >1.7.3,<1.8 || >1.8,<1.8.1 || >1.8.1,<2.0.0
 - linkml *
 - linkml-runtime >=1.10.2,<2.0.0
```

glass-ships:
Closes #260 

### Summary

- Updates Pydantic to v2.5, along with some other packages to allow this
- Updates fixtures, scripts, and migrate related code 
- Updates tests
  - curie expansion fails: on omim, ensembl, uberon, and wormbase (potentially unrelated?)


glass-ships:
curie expansion fails on: omim, ensembl, uberon, and wormbase (returns None)
@cthoyt do you by chance have an idea why this would be the case? 
related code: https://github.com/monarch-initiative/monarch-app/blob/pydantic-v2/backend/src/monarch_py/service/curie_service.py

cthoyt:
I can suggest you update the tests to use strict=True in the expand call then debug how prefix maps is constructing the EPM. This seems like a data and not a code issue

glass-ships:
interesting, yeah...
```
        prefix, identifier = self.parse_curie(curie)
        rv = self.expand_pair(prefix, identifier)
        if rv:
            return rv
        if strict:
>           raise ExpansionError(curie)
E           curies.api.ExpansionError: WormBase:WBGene00000001
```
not sure where to start digging to sus this out

cthoyt:
@glass-ships I wrote a minimal test that shows that there is no WormBase prefix available through the "merged" context in `prefixmaps` https://gist.github.com/cthoyt/ef44b41c11da966cfde322a9fac26859. 

Here's the situation:

1. The source file for the v0.2.0 release, [`merged.csv`](https://github.com/linkml/prefixmaps/blob/v0.2.0/src/prefixmaps/data/merged.csv), has several different related records Screenshot 2023-12-07 at 09 40 38 that differ in http vs. https, but this shouldn't cause the issue



2. On a first pass, I thought this was related to an issue with the merging algorithm (which is definitely apparent from this screenshot), but I don't think this is the case. Some context about that thought process:
   > I think there is an assumption that this situation does not happen - there should only be one "canonical" record with a given namespace. Therefore, this seems like an issue with the construction of the merged context.
   > - I previously observed a similar issue for Wikidata and reported it in https://github.com/linkml/prefixmaps/issues/49
   > - This could be a bug in the algorithm that makes the merged context, which I reported in https://github.com/linkml/prefixmaps/issues/50
   >
   > 
   > Discussion on https://github.com/linkml/prefixmaps/issues/26 is also relevant - there, I suggested to apply some data integrity testing to the merged contexts

3. On a second thought, this seems like an even weirder error since WormBase isn't getting loaded at all in the converter (see the notebook). I still think this is a data issue and is not coming from `curies`

glass-ships:
maybe a silly question but, when you say data error, what data in particular do you refer to?

cthoyt:
The issue is happening because the wrong data is inside the `curies.Converter` object, not because there is a bug in the `curies` code. This means we need to look in the `prefixmaps` code or data to figure out why the right stuff doesn't make it inside the object that you're actually interacting with

kevinschaper (approved):

PR #497, opened 2023-12-11
Bring back oak and phenio (text annotator)

### LINKED ISSUE #234: Update text annotation capabilities for Monarch App
Use OntoRunner/OGER to replace the old text annotator. 
https://github.com/monarch-initiative/ontorunner
https://github.com/OntoGene/OGER

The main workflow for the Monarch app is to enter text and extract phenotypes to be exported to the explorer. 

The old endpoint looked like: 
```
const url = ${biolink}/nlp/annotate/entities;
const response = await request<_Annotations>(url, params, options);
const { spans } = response;
```

vincerubinetti:
Closes #234


glass-ships:
@vincerubinetti @madanucd:

Madan has updated the backend to provide the proper format, tested locally and integrates with the frontend nicely, I'll add some fixture data tomorrow. 

One thing I noticed - the API doesn't know how to handle line breaks. 
Can (should) we strip newlines from input content before making the API request? 
Should the API be adjusted to keep newlines?

vincerubinetti:
For best UX, I think newlines should be preserved. Imagine uploading several separated paragraphs, then getting back a long blob of text. Harder to find the start of sections you want to look at.

I could do some work-arounds and trickery to remember the newlines and re-insert them after the response, but I think it'd be better to have the backend preserve the newlines (and any whitespace in general).

glass-ships:
yeah i think that makes sense too. sweet, we'll look at getting that functionality into the backend!

glass-ships:
image

So i'm ... closer? but clearly `<br>` was the wrong thing to use. @vincerubinetti  What should i have in `{ "text": ??? }` for newlines?

vincerubinetti:
`<br>` is the right tag, it's just that Vue (and any other framework) will escape HTML by default if you just pass it as text. You'd have to use `v-html` for that. But that sort of defeats the purpose of what we're trying to do here.

The results get shown in a `<p class="results">` which has `white-space: pre-line`, which means it will respect white space. So, you should just need to put new line `\n` characters there, which is also what should be being passed to the backend through the multi-line textbox.

vincerubinetti:
See line 604 in the text-annotator.json fixture in this branch for an example.

And the mock on the pr preview:

https://deploy-preview-497--monarch-app.netlify.app/explore?mock=true#text-annotator

kevinschaper (approved):

PR #511, opened 2023-12-15
remove old ssh keys in action

glass-ships:
Attempts to address issue with update-gcp action failing


falquaddoomi (approved):
Looks reasonable to me! Nothing blocking merging AFAICT.

One thing that might aid readability is adding in newlines; since you're using a YAML block (indicated by the `|` after `run: `) you should be able to use newlines just fine.

I also see that you opted for the version that doesn't filter by account, but I guess you only see the keys for the currently-activated account when you run `gcloud compute os-login ssh-keys list`? If that's the case, then I suppose it'll only get rid of its own keys, so that's probably fine.

PR #515, opened 2023-12-19
generate category enums by solr query

glass-ships:
Generate category enums via Solr query, use as args in API calls (to pretty up swagger docs and clarify what available options there are)


kevinschaper:
Fixes #516

kevinschaper (approved):

PR #529, opened 2024-01-03
Fix semsim search GET api method, add api tests

kevinschaper:
The actual fix is just swapping "category" for "group" in the semsim search GET request, but I also wanted to start on getting full test coverage of our api methods.


backend/src/monarch_py/api/association.py
```diff
@@ -1,4 +1,4 @@
-from typing import List, Optional, Union
+from typing import List, Union
```
kevinschaper, line 1: This change came in from auto formatting

backend/src/monarch_py/datamodels/category_enums.py
```diff
@@ -3,139 +3,149 @@
 
 class EntityCategory(Enum):
     """Entity categories"""
```
kevinschaper, line 5: All of the changes in this file are from auto formatting

glass-ships (approved):

PR #530, opened 2024-01-04
TSV for Downloads

### LINKED ISSUE #505: adding downloadable tsv file of propagated gene annotations
The monarch web site currently offers tsv files for gene-phenotype associations for the major model species, which is an amazing resource! The files I reference are found at 
https://data.monarchinitiative.org/latest/tsv/gene_associations/index.html

However I don't believe these gene annotations are propagated (i.e. annotations to more specific terms are then also annotated to any ancestor node). I was wondering if you all at Monarch could provide a file of that nature? The idea of this is to be able to do species specific gene set enrichment on a collection of phenotypes.

Thanks!

glass-ships:
Closes #505


glass-ships:
so i've added an API query parameter to allow fetching some items in TSV format, but I'm not sure how to adjust the frontend to be able to account for this.
@vincerubinetti could you point me in the right direction / would you be interested in taking point on that?

vincerubinetti:
This should be doable pretty quickly, we can do this on Friday unless there it is very urgent.

glass-ships:
nice! yeah i think that should be fine

kevinschaper (approved):

PR #540, opened 2024-01-16
Skip delete SSH keys when none are present

glass-ships:
Resolves issue with deleting GCP ssh keys if none are there


amc-corey-cox (approved):
I wrote this and Glass took a look. This should prevent the error when trying to remove SSH keys by skipping removal when no keys exist.

PR #545, opened 2024-01-20
add color to category icons

vincerubinetti:
- include tailwind color palette (professionally curated). manually pick from that, and assign a color to each category type.
- remove association icons
- remove hard-coded circle in category icons
- refactor icon component and simplify. add outline circle dynamically. make solid fill and white stroke inside for categories. tweak css.
- add slot to app tile component
- rework homepage to use two node icons for association tiles


glass-ships (approved):
looks great!

vincerubinetti:
image

Looks like the color is not being applied to some categories that I'm pretty sure we assigned colors to. Need to look into that.

glass-ships:
Oh interesting yeah, wonder if it's to do with it going with initials vs. having a dedicated icon. 

I also wonder if causal-disease / correlated-disease should also get the disease color

vincerubinetti:
My intent was to have those two things (initials and colors) basically be completely independent of each other. But I may have messed something up in the AppIcon component or related.

PR #563, opened 2024-01-24
fix download=false still downloading

glass-ships:
Resolves download=false still causing endpoint to download


kevinschaper:
It would probably be good to add a type hint on download in the function definition, I think then swagger will give a nice pulldown, and it's possible that the truthiness eval would work

glass-ships:
> It would probably be good to add a type hint on download in the function definition, I think then swagger will give a nice pulldown, and it's possible that the truthiness eval would work

added

kevinschaper (approved):

PR #565, opened 2024-01-25
Increase grounding precision

kevinschaper:
I pulled the grounding job out of spacy_implementation and gave it an interface that solr_implementation implements, and then changed the behavior by searching for quoted phrases against keyword fields. Also a quick fix of the text annotate GET endpoint for entities

### DIFF
--- a/.github/workflows/test-backend.yaml
+++ b/.github/workflows/test-backend.yaml
@@ -37,14 +37,14 @@ jobs:
         with:
           version: ">= 447.0.0"
 
-#      - name: "Cache SSH Key"
-#        uses: actions/cache@v3
-#        with:
-#          path: |
-#            ~/.ssh/google_compute_engine
-#            ~/.ssh/google_compute_engine.pub
-#          key: ${{ runner.os }}-ssh-key
-#
+      - name: "Cache SSH Key"
+        uses: actions/cache@v3
+        with:
+          path: |
+            ~/.ssh/google_compute_engine
+            ~/.ssh/google_compute_engine.pub
+          key: ${{ runner.os }}-ssh-key
+
 #      - name: "Remove old SSH keys"
 #        run: |
 #          if [[ $(gcloud compute os-login ssh-keys list | grep -v FINGERPRINT) ]]; then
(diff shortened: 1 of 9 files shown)


kevinschaper:
I think this ends up making grounding case sensitive, and I think we'll need another PR and a data update to add a lowercased keyword tokenized field in Solr.

madanucd:
<html xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:dt="uuid:C2F41010-65B3-11d1-A29F-00AA00C14882"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=OneNote.File>
<meta name=Generator content="Microsoft OneNote 15">
</head>

<body lang=en-US style='font-family:Calibri;font-size:11.0pt'>


<p style='margin:0in;font-family:Calibri;font-size:11.0pt'>Yes, we could
tighten up the results with exact match. However, some of the results yielded
provide reference to an exact match versus all the associated queries which
contain part of the string. Please see below a couple of examples comparing the
two results. We need to discuss on the best approach to address the generalized
string match issue.</p>

<p style='margin:0in;font-family:Calibri;font-size:11.0pt'>&nbsp;</p>

<div style='direction:ltr'>


Query | Result | Result (exact   match)
-- | -- | --
LEOPARD syndrome | LEOPARD syndrome 1  (MONDO: 0100082)      LEOPARD syndrome 2 (MONDO:         0012691)     LEOPARD syndrome 3 (MONDO:         0013380) | Noonan syndrome   with multiple lentigines (MONDO: 0007893)
Ocular albinism | Ocular albinism         (MONDO:0017304)     Ocular albinism         (HP:0001107)      Ocular albinism         (MP:0006159) | Ocular albinism   (MONDO:0017304)



</div>


</body>

</html>

kevinschaper:
I actually think the first example in that table is correct, since it's the parent to the three Leopard Syndrome subtypes and it's an exact match on a synonym (maybe in this case that synonym is a previous name?)

The second looks like it's the case sensitivity that needs a different fieldtype in Solr to solve, which I'd rather handle as a followup PR

madanucd (approved):
Looks good to me.

PR #569, opened 2024-01-30
Add button to run Text Annotator

### LINKED ISSUE #560: Text annotator: Add 'submit' button
## Overview
When I used the [Text Annotator](https://monarchinitiative.org/explore#text-annotator), to trigger submission of text and get annotation results the options appear to be: (a) click outside the text box, or (b) refresh the page.

This wasn't immediately clear to me and at first I thought it was broken. I tried refreshing the page first and it was only after playing with it for a while that I realized that clicking outside the box also works.

Can we add like a _Submit_ button so it is more clear how to use it?

An alternative might be to do auto-refresh. You could do this with a delay of maybe 1-2 seconds after the user stops typing. Could also add a small spinner/progress icon animation in some corner of the app while it is refreshing.

glass-ships:
Closes #560 

Considered adding an `<hr />` right above the annotator button, but I'll leave it to vince to make a visual design judgement there


frontend/src/pages/explore/TabTextAnnotator.vue
```diff
@@ -108,12 +111,6 @@ async function onUpload(data = "", file = "") {
   await runAnnotateText();
 }
 
-/** on textbox change */
-async function onChange() {
-  filename.value = "";
-  await runAnnotateText();
```
vincerubinetti, line 114: Let's keep the `onChange` call and just remove the `await line` (and then you can remove the `async` too. Resetting the `filename` variable is nice to keep. If someone uploads a file, it keeps track of that filename, and then if someone pasted a completely different text into the box, it would still show the old filename.

Otherwise, this looks good.

glass-ships: sounds good!

vincerubinetti (approved):

Issue #552, opened 2024-01-23
Include phenotype information on disease and gene nodes 


kevinschaper:
Related to #523, we want to have an autocomplete endpoint (or a boolean on the existing endpoint), we need to capture phenotype counts on disease and gene nodes in Solr. 

I think it might make the most sense to have a Golr-expansion like using the predicate name for the field value, and having IDs stored in the field named after just the predicate, then labels and counts as well for completeness, with negated phenotypes being excluded

```
has_phenotype: 
has_phenotype_label:
has_phenotype_count:
```

Issue #469, opened 2023-11-11
A query for `aggressive behavior` should return the HPO match above the GO match


kevinschaper:
This might be a somewhat tough result math compromise to get it working, but I searched for `aggressive behavior` hoping to get to HP:0000718, and I was surprised that it was the 9th result. 

There are disease matches coming in ahead, and I think it makes sense to give more boost-juice to disease than phenotypic feature, I think phenotypic features also need a bit of a boost over GO terms. 

There's likely something happening in the scoring math because of the case sensitivity difference, in that the GO (& NBO) aggressive behavior terms have exact match text, while the HPO term starts with a capital.

PR #581, opened 2024-02-07
has_phenotype expansion in node schema, semsim autocomplete and boosting tweaks

kevinschaper:
We initially/primarily only need has_phenotype_count, but once we're going to the trouble to join it's a shame to skip over the rest of the expansion


kevinschaper:
Fixes #552 and #469

amc-corey-cox (dismissed):
It looks good to me. I noticed Codecov flagged a few lines as not having coverage but they don't seem to be anything important. Did you change a few of the diseases used in some of the frontend test fixtures? Or is that just fuzz in the diff?

amc-corey-cox (approved):
Okay, I tested and it is working for me now. The problems I had were just my inexperience in testing this stack.

PR #583, opened 2024-02-07
Test coverage

amc-corey-cox:
### Related issues

- Testing coverage with Codecov in previous commit

Appears we need to add coverge.xml to main to get his to work. This PR now does this.


amc-corey-cox:
This just adds the coverge.xml report to main. Adding this should allow commits from here on to have coverage reports.

kevinschaper (approved):

PR #586, opened 2024-02-08
testing coverage

amc-corey-cox:
### Related issues

- Check to make sure coverage report is working.

### Checks


amc-corey-cox:
Not sure why this isn't working... I think we should sit on it a bit and see if it starts working. If not, I'll dig in again.

PR #592, opened 2024-02-09
Update checkout action

glass-ships:
Updates workflows to use checkout action v4, as v3 uses node 16 which will be deprecated soon.


glass-ships:
hey @vincerubinetti just double checking, this type check is expected to fail for now, right?

vincerubinetti:
No that shouldn't be failing.

I think this is a result of that quick merge conflict resolution we did earlier today with `srcs/pages/metadata.json`. If you look there now, there's no `icon2` fields. I can't fix this now, can you you go back in the file history and readd those changes:

```
https://github.com/monarch-initiative/monarch-app/pull/545/commits/a2f5df3276f5096b88a29df4cb7b6c83b2a24c6e
```

glass-ships:
yeah no worries! i'll take a look

PR #610, opened 2024-02-23
Turn off autoplay on ios video

### LINKED ISSUE #607: Google App iPhone autoplays video on mi.org
When opening monarchinitiative.org using Google App on iPhone, the videos on the homepage autoplay making it difficult to use to site. It does not seem to happen when opening Safari on iPhone. I have not checked it it happens using Chrome on iPhone.

vincerubinetti:
Closes #607

Note that this isn't a perfectly reliable solution but it's probably the best we can do.

### DIFF
--- a/frontend/src/components/AppHighlight.vue
+++ b/frontend/src/components/AppHighlight.vue
@@ -4,7 +4,7 @@
 
 <template>
   <div class="highlight">
-    <video :src="src" muted autoplay loop controls></video>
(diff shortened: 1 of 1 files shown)


glass-ships (approved):
Seems fine on ipad! 
Didn't autoplay at all on main or this branch's deploy preview. 
Still couldn't get `yarn dev --host` to work so i couldn't check the os locally, but it seems unaffected by this regardless, and doesn't have the same issue as iOS

PR #616, opened 2024-02-28
Add --traverse-orthologs flag for assoc table

### LINKED ISSUE #326: Will you still be able to access model organism phenotypes from human genes?
**Please describe your question, suggestion, or concern.**

On the current website from a human gene page you can access phenotypes of orthologs in model species. Will this feature be available on the new site?

v

If your question or suggestion is specific to Mondo, please submit it here instead: https://github.com/monarch-initiative/mondo/issues

glass-ships:
Closes #326 and #614

Flag runs without error, but still need to confirm.


kevinschaper (approved):

frontend/src/api/associations.ts
```diff
@@ -16,6 +17,7 @@ export const getAssociations = async (
     offset,
     limit,
     query: search || "",
+    traverse_orthologs: traverseOrthologs ? "true" : "false",
```
vincerubinetti: Just `traverseOrthologs` or `!!traverseOrthologs` should be sufficient here?

frontend/src/pages/node/AssociationsTable.vue
```diff
@@ -259,13 +261,14 @@ const {
     /** catch case where no association categories available */
     if (!props.node.association_counts.length)
       throw Error("No association info available");
-
+    console.log("Include Orthologs:", props.includeOrthologs.valueOf());
```
vincerubinetti: Remove before merge

frontend/src/pages/node/AssociationsTable.vue
```diff
@@ -259,13 +261,14 @@ const {
...
     /** get association data */
     const response = await getAssociations(
       props.node.id,
       props.category.id,
       start.value,
       perPage.value,
+      props.includeOrthologs.valueOf(),
```
vincerubinetti: This should just be `props.includeOrthologs`? I'm not sure what `valueOf` is doing there.

frontend/src/pages/node/AssociationsTable.vue
```diff
@@ -303,6 +306,10 @@ watch(
   () => props.category,
   async () => await queryAssociations(true),
 );
+watch(
+  () => props.includeOrthologs,
+  async () => await queryAssociations(true),
```
vincerubinetti, line 310: ~Technically don't need the `async` and `await` here since you're not waiting for it to finish to do something afterward.~

vincerubinetti:
Let's talk about this a bit tomorrow before merging. It's possible we can simplify things in the custom `useQuery` composable, and by using `watchEffect` more instead of just `watch`.

glass-ships:
sounds good, tomorrow works! there's no huge rush on the frontend aspect, as i still need @kevinschaper 's help sorting out some issues with the backend before merging anyway

thanks vince!

PR #619, opened 2024-03-01
Untruncate labels for phenogrid export

vincerubinetti:


glass-ships (approved):

vincerubinetti:
Is there a way to relax the code coverage, at least on the frontend. I think it's a bad idea to leave so many PRs with ❌s, as people will learn to ignore them (e.g. "ah it's probably just the code coverage failing"), and then you'll miss real problems.

glass-ships:
> Is there a way to relax the code coverage, at least on the frontend. I think it's a bad idea to leave so many PRs with ❌s, as people will learn to ignore them (e.g. "ah it's probably just the code coverage failing"), and then you'll miss real problems.

Yeah, I think @amc-corey-cox was gonna take a look at that, but we definitely need to

vincerubinetti:
@kevinschaper poke for review, unless you have questions which we can talk about tomorrow.

kevinschaper (approved):

PR #621, opened 2024-03-04
Sync url

### LINKED ISSUE #575: Sync more selections to url
It'd be good to sync things like search category filters, phenotype explorer options, etc. with the URL.

Integrate some kind of generic url-sync reactive variables, like this:
https://github.com/colorado-cancer-center/COCancerScope/blob/main/frontend/src/util/composables.ts

vincerubinetti:
Closes #575 

Still needs more testing before ready. Will _not_ hook up the node search using this because that is more complicated and manual.

Can add the LZW text-annotator sync to this PR too if you want.

- make "replace" param of link component manual
- split composables into their own files and folder
- add generic use-param composable that creates a two-way synced ref with a url param
- hook up use-param to phenotype explorer


vincerubinetti:
@kevinschaper poke for review, unless you have questions which we can talk about tomorrow.

kevinschaper (approved):

frontend/src/composables/use-floating.ts
```diff
@@ -7,10 +7,10 @@ import { useEventListener } from "@vueuse/core";
 /** use floating-ui to position dropdown */
 
 export const useFloating = (
-  anchor: Ref<HTMLElement>,
-  dropdown: Ref<HTMLElement>,
+  anchor: Ref<HTMLElement | undefined>,
+  dropdown: Ref<HTMLElement | undefined>,
   fit = false,
-) => {
+): { calculate: () => Promise<void>; style: Ref<CSSProperties> } => {
```
vincerubinetti, line 13: The explicit return type shouldn't be necessary here, ts should be able to infer it. If it is necessary, something is wrong.

monicacecilia:
:bowtie:

vincerubinetti:
A minimal testbed just for this "use-param" composable:

[vue-use-params.zip](https://github.com/monarch-initiative/monarch-app/files/14661743/vue-use-params.zip)

Updated posthumously to fix a bug we encountered, where when having a phenogrid iframe with multiple url params of the same name, this composable would redirect the iframe and drop some of the parameters, because it originally worked under the assumption that each key only appeared once.

PR #622, opened 2024-03-04
Set to Sets Phenogrid for IMPC

### LINKED ISSUE #599: Add multi-set Phenogrid for IMPC
IMPC is currently using a legacy.mi.org endpoint that takes one set of phenotypes and allows it to be compared against multiple other sets. As a first step towards being able to provide continuity, we should add that option to the Monarch API. 

My first reaction is that because the input would be close to compare, but not exactly the same, and the output would be close to search, but not exactly the same, we might want to just make a dedicated multi-compare endpoint. 

On the other hand, the notion of providing an overall score might be beneficial to compare, and so that would be a strong argument in favor of changing up our compare endpoint to take one `subjects` collection and many `objects` collections. Whether you have one or many, each `objects` collection would be given a score relative to the `subjects` set. A name and ID could also be passed in, so that the sets come out nicely annotated.

kevinschaper:
Adds a page, endpoint, and test-element in `/testbed` for a multi-compare phenogrid

It's possible there's a way to refactor the existing phenogrid page to be able to handle requests for either, and we can definitely go that route if @vincerubinetti has suggestions for how to do that. 

Just a rough first pass at implementing this functionality


glass-ships:
Something is wrong with the testbed iframes and I'm not sure what the issue is... 
Regular and multi-compare phenogrids seem swapped and one is rendering the full app instead of routing to where it's supposed to go

vincerubinetti:
Can we spend Friday on this instead of me making review comments/changes now?

glass-ships:
Yeah absolutely, I'm mainly just commenting/etc to keep track of where things are at. 
Friday sounds perfect

kevinschaper:
We'll keep working on it until then, but don't feel like you need to stay tuned in the whole time

glass-ships:
Just commenting to "checkpoint" progress. 
We may need to figure out how to POST for the iFrame, since the backend expects an `id` and `label` for each set of objects, but watching the url for those query params doesn't (conveniently) seem to offer those. 
That _should_ be the last thing we need to do for this PR though, hopefully something we can suss out on Friday

glass-ships:
__**DO NOT MERGE**__ 
Until @vincerubinetti has a chance to fix the "Sync URL to Parameter" issue. 
He'll let us know when he gets around to it

frontend/src/pages/PageTestbed.vue
```diff
@@ -10,16 +10,28 @@
...
   </AppSection>
 
+  <!-- phenogrid -->
+  <AppSection>
+    <AppHeading>Phenogrid - Multicompare mode</AppHeading>
+    <iframe
+      name="pheno-multi"
+      title="MultiCompare Phenogrid"
+      frameBorder="0"
+      src="/phenogrid-multi-compare?subjects=HP:0002616,HP:0001763,HP:0004944,HP:0010749,HP:0001533,HP:0002020,HP:0012450&object-sets=HP:0002616,HP:0001763,HP:0000767,HP:0000023,HP:0002108,HP:0000490,HP:0000545,HP:0100785,HP:0000268&object-sets=HP:0002616,HP:0001763,HP:0004944,HP:0010749,HP:0001533,HP:0002020,HP:0012450,HP:0003394,HP:0003771,HP:0012378,HP:0001278,HP:0002827,HP:0002829,HP:0002999,HP:0003010&object-sets=HP:0002616,HP:0001763,HP:0000767,HP:0000023,HP:0002108,HP:0000490,HP:0000545,HP:0100785,HP:0000268,HP:0001634,HP:0001653,HP:0001659,HP:0002360,HP:0003179,HP:0004970,HP:0005059,HP:0002705,HP:0012432,HP:0007800,HP:0001704"
```
kevinschaper, line 31: @glass-ships We should convert the multi-compare phenogrid example to message-sending mode rather than url-mode

vincerubinetti: I think both should be supported and tested. This can have a button to trigger a message send like the search phenogrid example above it.

kevinschaper: Could we do a standalone testbed page just for multi-compare phenogrid via message?  That's the use case that we want to clearly provide to our IMPC friends.

vincerubinetti: Creating a whole new route for that seems like overkill and a lot of duplication. I feel that the testbed code, and more importantly the readme code, clearly demonstrates how to use it.

However it looks like the phenogrid instructions were moved to `contributing.md` for some reason. I think there should be a `phenogrid.md` or something that clearly instructs how to use it via url and post message. And it could use more clarity, like strongly encouraging the message mode of operation for set to sets (to allow labels and such).

At most, I think we could have separate sections on the testbed for each mode of operation. Or just make message mode the default and have a button to test the url mode.

vincerubinetti: Also somehow my additions/changes to the readme here were overwritten

https://github.com/monarch-initiative/monarch-app/pull/622/commits/9c666f8cc5fe90d2ae27762294f374dc20fc6717

glass-ships: yeah we definitely missed a few things when we merged the documentation changes from main. 

I just re-added your changes to the section in what is now `contributing.md`

kevinschaper: A message mode example for multi-compare on testbed along with a phenogrid.md makes sense.

glass-ships: This may be a good opportunity to expand the top level `docs/` directory to include frontend documentation and place it there. Having three separate README and CONTRIBUTING files feels a little clunky, scattered, and repetitive in my opinion. 
But that can be a discussion held somewhere else.

glass-ships:
Ok @vincerubinetti I've updated the documentation to re-introduce the changes you made, as well as adding a section on the multi-compare mode, let me know if you'd like to see any changes!

vincerubinetti:
I've updated the phenogrid documentation and moved it to its own file.

I've also pushed what I THINK is a fix for that URL sync bug. I still need to test it out and look and it more closely tomorrow.

See this comment I added to the original sync-url PR for a minimal testbed just for the "use param" composable:

https://github.com/monarch-initiative/monarch-app/pull/621#issuecomment-2008617517

frontend/src/pages/PageTestbed.vue
vincerubinetti: I've updated the documentation and I think it's pretty clear how to use it.

In fact, we should be telling people to look at that documentation, and not our testbed. Our testbed should only be for our own testing purposes and not a place of record.

vincerubinetti: If you really want an in-situ demo of the phenogrid widget to show people, a better idea would actually be to make a JSFiddle or something like that. That would also prove (and test) that it can work on a different domain (without CORS issues).

vincerubinetti:
Okay I've tested my fixes and they work to my satisfaction. Please look at my recent commits and also re-test the url-syncing a little bit. Then feel free to merge.

vincerubinetti (approved):

glass-ships:
As of latest commit, phenogrids in testbed both mocking and using local solr/semsimian seem to work.  
I'll go ahead and merge this, thanks for all your help with this Vince!

PR #623, opened 2024-03-05
Add InstanceType for defineExpose'd components

### LINKED ISSUE #615: Use `InstanceType<typeof MyComponent>>` for to get `defineExpose` type
Rather than re-specifying the type in the consuming component. Low-hanging fruit.

https://github.com/vuejs/core/issues/4397#issuecomment-1035863245

@glass-ships you could also take a stab at this if you have some time, otherwise we can do it quickly on friday.

glass-ships:
Closes #615

### DIFF
--- a/frontend/src/components/AppButton.vue
+++ b/frontend/src/components/AppButton.vue
@@ -51,7 +51,7 @@ const props = withDefaults(defineProps<Props>(), {
 });
 
 /** element ref */
-const button = ref();
+const button = ref<HTMLButtonElement>();
 
 /** type of component to render */
 const component = computed(() => (props.to ? "AppLink" : "button"));
--- a/frontend/src/components/AppDetail.vue
+++ b/frontend/src/components/AppDetail.vue
@@ -4,7 +4,6 @@
 
 <template>
   <AppFlex
-    ref="cell"
     align-h="left"
     direction="col"
     gap="small"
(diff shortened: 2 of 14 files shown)


glass-ships:
I may have missed something, but it looked like Phenogrid was actually the only place where an "exposed" component was used with a ref

vincerubinetti:
There's definitely other places:

Screenshot 2024-03-04 at 10 13 00 PM

Somewhere in the code, there should be a use of these components with a `ref` attached, which needs to be annotated. If there isn't, then there's no reason for the `defineExpose` to exist. At a glance, it looks like AppTextbox uses a ref on AppInput like `const input = ref();`. If you hover over that ref, it's `any` (meaning ts is turned off for that variable), which isn't good.

glass-ships:
ohhhh interesting, so it can expose individual variables, not just the whole `AppWhatever` component? 
ok I'll take a closer look tomorrow, guess I misunderstood at first

glass-ships:
Ok taking another look now. so, in the examples of `button` or `input`, what should the `typeof` be? 
it's not super clear from the components they're defined in, as they just tend to be `const button = ref()` for example

vincerubinetti:
See my changes.

I was able to check if a `defineExpose` is even needed with this regex in vscode: `<AppButton([^>]|\n)*ref([^>]|\n)*>`

vincerubinetti (approved):

glass-ships:
Gotcha, thanks for taking a look! I tried but I guess I still have some gaps to fill in

PR #624, opened 2024-03-05
Node/Yarn to Bun

vincerubinetti:
Converts repo from using Node and Yarn to just using Bun, and upgrades packages (notably Vite -> v5).

https://twitter.com/bunjavascript/status/1751726306793046490

This is not absolutely necessary. In my testing, it saves some headache and time, but it's also possible it could cause a problem down the road. See the added note in the readme.

I think using Bun just as a package manager is very low risk, so at the very least we could replace Yarn with Bun. In my testing, starting from cleared caches, Yarn took 1 minute to install the frontend dependencies, and Bun took 6 seconds (where simply downloading the packages was the main bottleneck). This could save us time waiting for frontend tests to finish, and possibly some $$ if we were to ever go over GitHub Actions quota. 

Using Bun as a Node replacement, again from anecdotal evidence, is also fairly low risk. They seem to have already covered the general, commonly used Node functionality, and have tested popular tools like Vite to make sure they work with Bun. In some edge cases though, Bun might not behave _exactly_ as Node. But that is a bug in Bun, as they explicitly say that anything in Node should also work in Bun. The other reason it's low risk is that it shouldn't be difficult to go back to Node if there is ever a problem. The biggest risk, I think, is there being a problem, and the maintainers not being aware that it could be a Bun bug.

Using Bun as a bundler or test runner would probably be a bigger change, and could cause more problems, in my estimation, so that is left out of this PR.

Another caveat: Bun's support of Windows is still in progress. They say about 90% of Bun's test suite passes on Windows.

@kevinschaper  If someone has a laptop that does NOT have node or yarn installed on it, I'd be interested to test to make sure that _just_ installing Bun will work as expected. Theoretically, Bun should be the only installation requirement for the frontend after this.


glass-ships (approved):
nice, looks good!

glass-ships:
it's weird to me that tests were working before i removed the node install step... and that we need a setup bun step for every job 

but hey, if it works i guess

vincerubinetti:
Ah yes, the e2e tests will be failing because of a quite stupid problem between playwright and linting:

https://github.com/microsoft/playwright/issues/23662
https://github.com/trivago/prettier-plugin-sort-imports/issues/270

Still working on a workaround for that. This is not due to Bun though, this is ultimately a result of upgrading to Vite 5 and it warning that `"type": "module"` in `package.json` will eventually be required, and then adding that is causing the above errors.

vincerubinetti:
> it's weird to me that tests were working before i removed the node install step... and that we need a setup bun step for every job
> 
> but hey, if it works i guess

Setup Node wasn't needed on all those steps before because the environments that GitHub Actions run come pre-installed with Node (along with other common things like Python and Ruby I think?). They do not come with Bun... yet. I predict that they will come with it in the next year or so. So eventually all those setup steps will not be necessary.

vincerubinetti:
Poking @kevinschaper  for review, especially regarding my fresh laptop comment above.

glass-ships:
re: fresh laptop, I spun up a fresh, barebones Ubuntu Docker container and ran the following:
```
apt update -y && apt upgrade -y && \
apt install git curl unzip build-essential -y && \
curl -fsSL https://bun.sh/install | bash && source /root/.bashrc && \
cd home && git clone https://github.com/monarch-initiative/monarch-app && \
cd monarch-app && git checkout bun && make install-frontend && \
cd frontend && bun run dev
``` 
and that did in fact seem to work without error. I couldn't remember how to expose a port with a Docker container, so unfortunately I wasn't able to actually _view_ the site. but i did get:
```
$ vite

  VITE v5.1.4  ready in 477 ms

  ➜  Local:   http://localhost:5173/
  ➜  Network: use --host to expose
  ➜  press h + enter to show help
```
which more or less confirms that just installing bun should be sufficient

vincerubinetti:
Nice, thanks for doing that. We can chat more about this tomorrow before merging.

amc-corey-cox:
nvm can be used for version managing in Node, does Bun have a similar feature? I know that javascript tends to lean towards a model of staying towards the newest versions but it is nice to have that kind of control when things break. How does Bun support this?

vincerubinetti:
That is a good question @amc-corey-cox , from some research it looks like it Bun doesn't have a "version manager" yet, either [built-in](https://github.com/oven-sh/bun/issues/3917) or [third-party](https://github.com/nvm-sh/nvm/issues/3189). Personally I am using `v1.0.0`, atm, which can be checked with `bun --version` like normal. We could specify a minimum or specific Bun version number in the readme, then the developer would (for now) have to go get that version manually somehow.

vincerubinetti:
We can shelve this for now, if we want. I'm unsure if the benefit it provides outweighs the cost/risk (which is still minimal imo). Locally, Bun is much much faster, for me. But on other people's laptops, it may not be.

Also I might be doing something non-optimally.  The cache step might be unnecessary anymore. I remember seeing some tweet (that I can't find now) by the Bun guy that the setup-bun gh action didn't cache anything because the time it took to store/restore cache actually took longer than bun just doing it without cache.

vincerubinetti:
https://github.com/oven-sh/setup-bun/issues/14#issuecomment-1714116221

vincerubinetti:
I just removed the cache step completely, and it runs significantly faster somehow:

https://github.com/monarch-initiative/monarch-app/actions/runs/8209866962 (cache miss, need to do cache-save before cache-restores)
https://github.com/monarch-initiative/monarch-app/actions/runs/8209901871 (cache hit, only need to do cache-restores)
vs.
https://github.com/monarch-initiative/monarch-app/actions/runs/8209932865 (no cache at all)

The time it takes to save node_modules cache (10s) plus the time it takes to restore it in each step (10s) takes WAY longer than just install it on every step.

I don't know how Bun is doing an complete install in less than a second... I'm going to ask in the thread I linked above how that's possible. But the proof is in the pudding: it's working.

glass-ships:
That's incredible, holy crap! 

Also re: installing specific version of bun, it's as easy as:
```
curl -fsSL https://bun.sh/install | bash -s "bun-v1.0.0"
```

for example, according to the docs

kevinschaper (approved):

vincerubinetti:
Forgot to remove `CACHE_PATH` as part of this.

PR #627, opened 2024-03-06
Rename ortholog phenotype checkbox

kevinschaper:
Removes option to traverse orthologs for disease associations, since there's no path to get to it right now. Updates label to be ortholog phenotypes to match naming in V2 ui.


glass-ships (approved):

kevinschaper:
thanks for the formatting cleanup!

PR #630, opened 2024-03-08
Add histopheno bar graph to gene pages

kevinschaper:
Use UPHENO to drive the Histopheno bar chart, then show the visualization on gene pages (likely limited to a subset of species, due to both the content of the list and the presence or absense of upheno connectivity)


kevinschaper:
@matentzn There were a few of these HP terms that we don't have UPHENO superclasses for. Do you have thoughts on that?  Even using UPHENO, maybe the lists still need to differ somewhat by species?

matentzn:
This is only possible if phenio imports uPheno AND HPO. Does it? If so that's very wrong..

caufieldjh:
> This is only possible if phenio imports uPheno AND HPO. Does it? If so that's very wrong..

I can confirm that PHENIO does not do that.

caufieldjh:
@kevinschaper Which HP terms are missing UPHENO superclasses?

kevinschaper:
@caufieldjh 

These are the two that I didn't find a direct UPHENO superclass for: 
```
    metabolism_homeostasis = "HP:0001939" # ???
    neoplasm = "HP:0002664" # ???
```

I haven't yet looked at what kind of coverage there is down from these upheno terms to MP/ZP etc

matentzn:
There are superclasses though: https://ols.monarchinitiative.org/ontologies/upheno2/terms?iri=http%3A%2F%2Fpurl.obolibrary.org%2Fobo%2FHP_0002664, https://ols.monarchinitiative.org/ontologies/upheno2/terms?iri=http%3A%2F%2Fpurl.obolibrary.org%2Fobo%2FHP_0001939

They are both under UPHENO:phenotype (and even UPHENO:phenotypeBySource) - Its true that there are many phenotypes for which there is no specific grouping, like UPHENO:neoplasm or some such..

glass-ships (approved):

PR #631, opened 2024-03-08
Multicompare API update

kevinschaper:
Convert the back end to doing a post for multi-compare and returning SemsimSearchResults so that it can use the same phenogrid as search


kevinschaper:
We should use `profile` in the model to generically refer to the list of IDs used in a compare

PR #635, opened 2024-03-14
Add option to select metric for semsim search/compare

### LINKED ISSUE #593: Add option to supply a metric for phenotype comparative search
Hello,

Our application is using 3 main API calls.
The first one is easy to migrate, but for the second and third, we are not sure what's the successor of this call. Can you please assist and tell us what new API call we should use?

https://api-biolink.monarchinitiative.org/api/bioentity/phenotype/HP:0012773

https://api-biolink.monarchinitiative.org/api/sim/search?taxon=9606&metric=phenodigm&limit=20&&id=HP:0012773
* note: metric can be 'phenodigm, jaccard, simGIC, resnik or symmetric_resnik'

https://api-biolink.monarchinitiative.org/api/bioentity/disease/MONDO:0014032/genes?use_compact_associations=true

glass-ships:
Closes #593 

- Adds an option to select from jaccard, phenodigm, and ancestor IC type scoring from semsim search and compare methods in: 
  - CLI
  - API endpoints
  - UI for phenotype explorer
- Updates backend tests
- Also updates frontend api method for calling semsim methods to include and use ancestor IC rather than top level score

I probably should have split front and backend into separate PRs so that the backend changes could be deployed and the frontend deploy previews would actually reflect changes being made. 

We can review this in hackathon tomorrow (3/15) before merging if we want to be sure, I'll screenshare to show it in action. 

Also somehow just noticed phenogrid rows are randomly sorted, just commenting so I remember to bring it up tomorrow. Already made a ticket #636

### DIFF
--- a/backend/src/monarch_py/api/additional_models.py
+++ b/backend/src/monarch_py/api/additional_models.py
@@ -19,6 +19,16 @@ class OutputFormat(str, Enum):
     tsv = "tsv"
 
 
+class SemsimMetric(str, Enum):
+    ANCESTOR_INFORMATION_CONTENT = "ancestor_information_content"
+    COSINE_SIMILARITY = "cosine_similarity"  # Not implemented
+    JACCARD_SIMILARITY = "jaccard_similarity"
+    PHENODIGM_SCORE = "phenodigm_score"
+
+    def __str__(self):
+        return self.value
+
+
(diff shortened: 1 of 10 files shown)


kevinschaper (approved):
yessss!

PR #637, opened 2024-03-15
AppRing updates

glass-ships:
Updates the AppRing component and some calculations to reflect ranges for different similarity metrics

Also sets angle to a full 360 if percent comes in as 1 (not working in preview for some reason)


glass-ships:
Locally i get this:
image

But the preview is showing an empty circle for 1.0 jaccard instead, which is... weird.

vincerubinetti (approved):

Issue #328, opened 2023-09-11
Some of the "top 5 phenotypes" are duplicates: plus repeating data type every time.


ValWood:
Some of the "top 5 phenotypes" are duplicates:

<img width="932" alt="Screenshot 2023-09-11 at 19 08 33" src="https://github.com/monarch-initiative/helpdesk/assets/7359272/0c788d7d-d9cf-46bf-8480-f57db0c15413">

That's a lot of space to say :
"[Aggressive behavior (HPO)] evidence (2)

Is it necessary to repeat the gene every time when you are on the gene page?

glass-ships:
https://monarchinitiative.org/HGNC:555#associations

kevinschaper:
It looks like these are separate lines in the HPOA genes_to_disease file based on the disease_id column: 

```
ncbi_gene_id	gene_symbol	hpo_id	hpo_name	frequency	disease_id
164	AP1G1	HP:0000718	Aggressive behavior	2/3	OMIM:619548
164	AP1G1	HP:0000718	Aggressive behavior	4/8	OMIM:619467
```

We should find a way to represent that in the UI. Maybe there's an appropriate qualifier or evidence field

kevinschaper:
Issue #430 is meant to generally address the idea of collapsing/grouping edges, but even if that were implemented, this particular example would still end up looking odd. 

The only difference between the two rows in the incoming file are frequency counts, so we want to be able to show that.

kevinschaper:
I'm going to move this forward to November and assign it to myself.  Another related bug/new feature that I'm seeing is that we aren't including frequency when it's represented by an HP term, like: 

```
ncbi_gene_id	gene_symbol	hpo_id	hpo_name	frequency	disease_id
113179	ADAT3	HP:0000718	Aggressive behavior	HP:0040284	OMIM:615286
113179	ADAT3	HP:0000718	Aggressive behavior	HP:0040283	ORPHA:363528
```

ValWood:
Personally, I think it could be a bit misleading to log the mentions in different resources as different evidences because quite often this will come from the same source. 

here the 2 sources are equivalent, this phenotype is associated with the disease description.
https://rarediseases.oscar.ncsu.edu/disease/intellectual-disability-strabismus-syndrome/about/

Using the current method will  begin to represent the number of resources, rather than the support for the phenotypic observation.

I probably didn't explain this very well. But I'm not sure this is what I would expect "frequency" to mean

CC @cmungall

[...]

PR #641, opened 2024-03-19
Add frequency column to phenotype pages

glass-ships:
Addresses part of #328


glass-ships:
Worth noting that tables are not very pretty, seemingly a direct result of adding more columns to a width limited component

Adds horizontal scroll bar and bunches up the name column

image

kevinschaper (approved):

PR #657, opened 2024-03-26
AppRing -> AppPercentage component

glass-ships:
This PR converts the AppRing component into an AppPercentage component with an optional bar type. 

This can be used in the AssociationsTable for frequency.

### DIFF
--- a/frontend/src/components/AppPercentage.vue
+++ b/frontend/src/components/AppPercentage.vue
@@ -3,30 +3,32 @@
 -->
 
 <template>
-  <div v-if="type === 'ring'" class="ring">
-    <svg viewBox="-50 -50 100 100">
-      <circle class="back" cx="0" cy="0" r="50" />
-      <path class="fill" :d="arc" />
-    </svg>
-    <b>
-      <slot />
-    </b>
-  </div>
+  <div v-tooltip="props.tooltip">
+    <div v-if="type === 'ring'" class="ring">
+      <svg viewBox="-50 -50 100 100">
+        <circle class="back" cx="0" cy="0" r="50" />
+        <path class="fill" :d="arc" />
+      </svg>
+      <b>
+        <slot />
+      </b>
+    </div>
 
-  <div v-if="type === 'bar'" class="bar">
-    <b>
-      <slot />
-    </b>
-    <svg viewBox="0 0 70 10">
-      <rect class="back" x="0" y="0" width="100%" height="100%" />
-      <rect
-        class="fill"
-        x="0"
-        y="0"
-        :width="`${100 * percent}%`"
-        height="100%"
-      />
-    </svg>
+    <div v-if="type === 'bar'" class="bar">
+      <b>
+        <slot />
+      </b>
+      <svg viewBox="0 0 70 10">
+        <rect class="back" x="0" y="0" width="100%" height="100%" />
+        <rect
+          class="fill"
+          x="0"
+          y="0"
+          :width="`${100 * percent}%`"
+          height="100%"
+        />
+      </svg>
+    </div>
   </div>
 </template>
 
@@ -40,6 +42,8 @@ type Props = {
   percent?: number;
   /** design */
   type?: "ring" | "bar";
+  /** tooltip on hover */
+  tooltip?: string;
 };
 
 const props = withDefaults(defineProps<Props>(), {
--- a/frontend/src/components/AppPercentage.vue
+++ b/frontend/src/components/AppPercentage.vue
@@ -3,7 +3,7 @@
 -->
 
 <template>
-  <div v-tooltip="tooltip">
+  <div v-tooltip="tooltip" role="tooltip" aria-hidden="true">
     <div v-if="type === 'ring'" class="ring">
       <svg viewBox="-50 -50 100 100">
         <circle class="back" cx="0" cy="0" r="50" />
--- a/frontend/src/components/AppPercentage.vue
+++ b/frontend/src/components/AppPercentage.vue
@@ -10,25 +10,18 @@
       <path :d="d" />
     </svg>
   </div>
-  
+
   <div v-else class="bar">
     <svg viewBox="-50 -50 100 30">
+      <rect x="-50" y="-50" rx="3" width="100%" height="100%" fill="none" />
       <rect
-      x="-50"
-      y="-50"
-      rx="3"
-      width="100%"
-      height="100%"
-      fill="none"
-      />
-      <rect
-      x="-49"
-      y="-49"
-      rx="3"
-      :width="fillWidth"
-      height="95%"
-      fill="none"
-      class="fill"
+        x="-49"
+        y="-49"
+        rx="3"
+        :width="fillWidth"
+        height="95%"
+        fill="none"
+        class="fill"
       />
     </svg>
     <b>{{ score.toFixed(1) }}</b>
@@ -63,7 +56,7 @@ const d = computed(() => {
   return `M 0 -50 A 50 50 0 ${angle >= 180 ? 1 : 0} 1 ${x} ${y}`;
 });
 
-const fillWidth = computed(() => `${clamp(props.percent*100, 0, 98)}%`);
+const fillWidth = computed(() => `${clamp(props.percent * 100, 0, 98)}%`);
 </script>
 
 <style lang="scss" scoped>
@@ -123,7 +116,8 @@ const fillWidth = computed(() => `${clamp(props.percent*100, 0, 98)}%`);
     stroke: $light-gray;
     stroke-width: 3px;
     &.fill {
-      fill: $theme;
+      // fill: $theme;
+      fill: $theme-mid;
       stroke: none;
     }
   }
--- a/frontend/src/components/AppPercentage.vue
+++ b/frontend/src/components/AppPercentage.vue
@@ -3,7 +3,13 @@
 -->
 
 <template>
-  <div :class="type">
+  <div
+    v-tooltip="tooltip"
+    :class="type"
+    tabindex="0"
+    role="button"
+    aria-label="Percent"
+  >
     <svg v-if="type === 'ring'" viewBox="-50 -50 100 100">
       <circle class="back" cx="0" cy="0" r="50" />
       <path class="fill" :d="arc" />
@@ -36,12 +42,15 @@ type Props = {
   percent?: number;
   /** design */
   type?: "ring" | "bar";
+  /** description of what percent represents */
+  tooltip?: string;
 };
 
 const props = withDefaults(defineProps<Props>(), {
   score: 0.5,
   percent: 0.5,
   type: "ring",
+  tooltip: undefined,
 });
 
 /** arc svg path */
--- a/frontend/src/components/AppPercentage.vue
+++ b/frontend/src/components/AppPercentage.vue
@@ -24,7 +24,7 @@
         class="fill"
       />
     </svg>
-    <b>{{ `${(score * 100).toFixed(1)}%` }}</b>
+    <b>{{ `${(score * 100).toFixed(0)}%` }}</b>
   </div>
 </template>
 
--- a/frontend/src/components/AppPercentage.vue
+++ b/frontend/src/components/AppPercentage.vue
@@ -3,12 +3,35 @@
 -->
 
 <template>
-  <div class="ring">
-    <div>{{ score.toFixed(1) }}</div>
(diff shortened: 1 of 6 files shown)


frontend/src/components/AppPercentage.vue
```diff
@@ -3,13 +3,29 @@
 -->
 
 <template>
-  <div class="ring">
-    <div>{{ score.toFixed(1) }}</div>
+  <div v-if="props.type === `ring`" class="ring">
```
vincerubinetti: `props.` is never needed within templates, vue unwraps this for us.

frontend/src/components/AppPercentage.vue
```diff
@@ -3,13 +3,29 @@
...
+    <b>{{ score.toFixed(1) }}</b>
     <svg viewBox="-50 -50 100 100">
       <circle cx="0" cy="0" r="50" />
       <path :d="d" />
     </svg>
   </div>
+
+  <div v-else class="bar">
+    <svg viewBox="-50 -50 100 30">
+      <rect x="-50" y="-50" rx="3" width="100%" height="100%" fill="none" />
```
vincerubinetti: If you were doing this to have the fill on top of the stroke, you can just use a single `<rect>` and use `paint-order: stroke`. However, I think it looks better to just have a gray background fill underneath, like the ring mode currently has.

frontend/src/components/AppPercentage.vue
```diff
@@ -3,13 +3,29 @@
...
+        x="-49"
+        y="-49"
+        rx="3"
+        :width="fillWidth"
+        height="95%"
+        fill="none"
+        class="fill"
+      />
+    </svg>
+    <b>{{ `${(score * 100).toFixed(0)}%` }}</b>
```
vincerubinetti: Since we have a bunch of different metrics/measures and places where we might use this component, I'm envisioning it's probably better to just get rid of the `score` prop all together and just use a `<slot>` here, and let the consuming component define number of decimal places, "%", etc. This way we don't end up with a hell of too many enumerated options for the component.

frontend/src/components/AppPercentage.vue
```diff
@@ -37,6 +55,8 @@ const d = computed(() => {
   const y = -cos(angle) * 50;
   return `M 0 -50 A 50 50 0 ${angle >= 180 ? 1 : 0} 1 ${x} ${y}`;
 });
+
```
vincerubinetti, line 64: I don't think this needs to be clamped at all. Also, nitpicking, but when the computation is this simple, and not repeated anywhere, I feel you might as well just put it directly in the `<template>`, no need for `computed` property. Vue style guides do recommend that you use computed properties as much as possible to keep the template simple, but I feel this can really be taken to an extreme.

frontend/src/components/AppPercentage.vue
```diff
@@ -71,4 +91,35 @@ const d = computed(() => {
...
+    height: 100%;
+  }
+
+  rect {
+    fill: none;
+    stroke: $light-gray;
+    stroke-width: 3px;
+    &.fill {
+      // fill: $theme;
+      fill: $theme-mid;
```
vincerubinetti: There is probably not enough contrast between `$theme-mid` and the white background or gray outline. Instead, you should probably use the darker `$theme`, as the ring mode does, and just put the text % above or to the side. It's not the most ideal for saving space, but it would be the most accessible.

If you wanted to keep the text on top of the bar and use `$theme`, you'd have to make use of `mix-blend-mode` to make sure any portion of the text over `$theme` is white and any portion over white/light gray is black.

frontend/src/pages/node/AssociationsTable.vue
```diff
@@ -332,6 +327,29 @@ async function download() {
...
+        return 0.3;
+      case "HP:0040283":
+        return 0.05;
+      case "HP:0040284":
+        return 0.01;
+      default:
+        return 0.0; // Default value if HP ID not found
+    }
+  }
+  return undefined;
```
vincerubinetti: Functions return undefined by default.

frontend/src/pages/node/AssociationsTable.vue
```diff
@@ -332,6 +327,29 @@ async function download() {
...
 }
 
+const frequencyPercentage = (row: DirectionalAssociation) => {
+  // Returns decimal, convert to percentage in template
+  if (row.has_percentage) return row.has_percentage / 100;
+  else if (row.has_count && row.has_total) return row.has_count / row.has_total;
+  else if (row.frequency_qualifier) {
+    switch (row.frequency_qualifier) {
+      case "HP:0040280":
+        return 1.0;
```
vincerubinetti: Decimal points on integers aren't necessary because all numbers in js are floats anyway.

vincerubinetti (commented):
I've made some comments, and committed some suggestions addressing them.

glass-ships:
Nice, thanks for these updates! 

I think this works nicely with the main use case being the associations table. 

Yeah the only reason I'd clamped the value was so the inner bar fit nicely within the outer box when full, but that's totally not necessary with the new design (and probably even resolved by the `paint-order` suggestion).

I'm tracking with most of your comments, but one thing I'm curious about is the single `rect` with the `paint-order: stroke` attribute. How would this work when the fill isn't the full width (ie. 100%) of the svg? Wouldn't we still need to have a second `rect` for the fill amount and specify the width?

vincerubinetti:
Sorry you're right paint order wouldn't help if you needed different widths for the rects

glass-ships:
Actually one more question (and no pressure responding, I know you're recovering!)

But I noticed we're able to have a tooltip in `TabPhenotypeExplorer` on the `AppPercentage` component, but when I try in the associations table adding:
```ts
      <AppPercentage
        v-if="frequencyPercentage(row) !== undefined"
        v-tooltip="'test tooltip'"
        :percent="frequencyPercentage(row)"
        type="bar"
        >{{
          ((frequencyPercentage(row) || 0) * 100).toFixed(0)
        }}%</AppPercentage
      >
```
nothing seems to happen... Same thing in the testbed. 
Is there anything special needed to enable tooltips in a component? or...

vincerubinetti:
I think this is because of this:
https://vuejs.org/guide/components/attrs

Since the component doesn't have a single root, vue doesn't automatically transfer the extra props (besides the ones defined in our `Props`) to the root elements in the component template. You can get around this by manually doing `v-bind="$attrs"` on both root elements, but I think more safe and consistent with the way I've done it elsewhere in the codebase, simply make `tooltip` a new explicit `Prop`, then pass it through to `v-tooltip` on both root elements.

`v-bind="$attrs"` is sort of like doing this in react: `function SomeComponent({ text, ...props }) { return <div {...props}>{text}</div> }`

glass-ships:
Oh iiiinteresting, gotcha! Thanks, I think I've worked something out - will push shortly

glass-ships:
Ok got it! I've removed the label and added an arrow function to make a tooltip based on how the frequency was calculated. 

I tried to make it similar to the phenotype compare "match" tooltip which is hover-over-able and includes an AppNodeBadge, but barring that, simply showing the frequency qualifier label is good enough

Edit: Oh god, whatever I did really messed with the e2e accessibility test

Update: added `role="tooltip" aria-hidden="true"` to outer div in AppPercentage component. Fixes e2e test. Though it's possible we don't want aria-hidden to be true here so that screen-readers can see it, but I wasn't sure what we'd then want the label to be

frontend/src/components/AppPercentage.vue
glass-ships: I did try and play with this for a while, but unfortunately most blending modes available didn't really work out (at best, it could be black over grey, and orange over the teal...)

Will probably just leave it as text off to the left.

frontend/src/components/AppPercentage.vue
```diff
@@ -3,7 +3,7 @@
 -->
 
 <template>
-  <div v-tooltip="tooltip">
+  <div v-tooltip="tooltip" role="tooltip" aria-hidden="true">
```
vincerubinetti: A few things about this:

- Role is meant to give accessible meaning to an element. Aria-hidden hides an element from the accessibility tree. As such, this really serves to quiet the error and bypass the test. Aria-hidden is appropriate when something is presentational, but this tooltip is not.
- The tooltip role is meant for the box that is actually the content of the tooltip, not the trigger. The reason the e2e axe tests are complaining is that div has no semantic meaning and is also not inherently interactive (e.g. can't tab to it). The way we've dealt with this in other places in the app (like the text annotator, where the results are just spans that have tooltips) is to assign `tabindex="0"` so you can tab to them and a person using a screen reader could access the tooltip.
- I strongly recommend always minimizing the number of elements you put in the DOM, to minimize "div hell" and css positioning and selector complexity. Here you have a div simply to avoid putting the same `v-tooltip` on two elements, which there's a different way to avoid.

glass-ships: Gotcha yeah, was definitely not sure about that "fix", I appreciate the detailed feedback! 

So step one would be replacing the `aria-hidden` with `tabindex="0"`. 

And I can get rid of the unnecessary outer div, but what would be the way to avoid adding `v-tooltip="tooltip" tabindex="0"` on both the inner divs?

Guess I could move the `v-if` to the different svg elements and replace `class="ring/bar"` with `:class="type"`, unless you had something else in mind

vincerubinetti: In the past for things like this I've used a role of `button`, but perhaps `note` or `definition` would actually be better, I don't really know. Here's an SO post on the subject:

https://stackoverflow.com/questions/70742579/what-is-the-correct-aria-role-for-a-more-info-icon

We don't want role `img` or `presentational` or anything like that because I believe then screen readers would ignore reading the number text as well.

In fact, just using `<meter>` semantically here might even be better, but then we also have to give it min and max values which could be a pain (or useful).

vincerubinetti: > what would be the way to avoid adding v-tooltip="tooltip" tabindex="0" on both the inner divs?

I'd honestly recommend duplicating some props in your template over duplicating actual divs in the browser. However in this particular case there was an easy way around it. If you had a situation when you really needed two root elements and needed to pass a lot of the same props to both of them, you could just group them together down in the `<script>` and pass them in via `v-bind`.

glass-ships: Gotcha, that makes sense, so something like [this SO post](https://stackoverflow.com/a/46731340). 

I was also just about to push a change I made locally before I noticed your commit. 
Would this also be a viable approach, or is it code-smelly for its own reasons? 
```ts
<template>
  <div :class="type" v-tooltip="tooltip" tabindex="0" role="tooltip">
    <svg v-if="type === 'ring'" viewBox="-50 -50 100 100">
      <circle class="back" cx="0" cy="0" r="50" />
      <path class="fill" :d="arc" />
    </svg>
    <b>
      <slot />
    </b>
    <svg v-if="type === 'bar'" viewBox="0 0 70 10">
      <rect class="back" x="0" y="0" width="100%" height="100%" />
      <rect
        class="fill"
        x="0"
        y="0"
        :width="`${100 * percent}%`"
        height="100%"
      />
    </svg>
  </div>
</template>
```

vincerubinetti: That looks great, and gets around needing to use the `<template v-if`'s in my commit. You can change it to that if you want.

It's still `role="tooltip"` in that snippet though, it should be button or something else.

I'm experimenting with role meter at the moment.

glass-ships: Oh right right, ok. Sounds good, thanks!

glass-ships:
I feel like I tried just adding `v-tooltip` in the component definition itself and it just didn't work... What the heck! maybe it's that bit with the type Slots...defineSlots()? Weird haha. 

This is looking great though, thanks for jumping in and helping out!

Out of curiosity, why pull all the tabindex/role/etc. stuff out of the component into a wrapping div when called?

vincerubinetti:
Sorry for the confusion. I've pushed hopefully the last commit, and it's now how I think it should be.

We can discuss what the difficulties were when we meet next time. They have to do with some peculiarities of how tippy.js applies aria attributes, and how we sometimes have the tooltip as a directive (`v-tooltip`) and as a slot (`<tooltip><template #content>`).

And before that most recent commit, the fact that sometimes the tooltip was on the ring itself and sometimes it was on a parent div was why I had the `tabindex` and `role="button"`s pulled out to the parents. But I simplified it.

PR #733, opened 2024-06-03
Show association selection as association category

kevinschaper:
Change up association table category count outputs to be based on the association category, rather than counterpart category. This can potentially be simplified by just generating labels from the association categories, but for now this PR is going to make a smaller change. Downstream, the UI needs a minor shift in the "sentence" structure of the association table heading/controls.


kevinschaper:
@glass-ships I think this is all good other than codecov complaining that the tests don't have test coverage.

backend/src/monarch_py/implementations/solr/solr_query_utils.py
```diff
@@ -52,14 +52,11 @@ def build_association_query(
             query.add_filter_query(" OR ".join([f'object:"{o}" OR object_closure:"{o}"' for o in object]))
     if entity:
         if direct:
-            query.add_filter_query(" OR ".join([f'subject:"{escape(e)}" OR object:"{escape(e)}"' for e in entity]))
+            query.add_filter_query(" OR ".join([f'subject:"{e}" OR object:"{e}"' for e in entity]))
```
kevinschaper, line 55: When I was initially debugging this problem, I noticed that we were escaping and quoting, I thought it made sense change this up to just quote, like we do most other places.

backend/src/monarch_py/implementations/solr/solr_query_utils.py
```diff
@@ -52,14 +52,11 @@ def build_association_query(
...
         else:
             query.add_filter_query(
                 " OR ".join(
-                    [
-                        f'subject:"{escape(e)}" OR subject_closure:"{escape(e)}" OR object:"{escape(e)}" OR object_closure:"{escape(e)}"'
-                        for e in entity
-                    ]
+                    [f'subject:"{e}" OR subject_closure:"{e}" OR object:"{e}" OR object_closure:"{e}"' for e in entity]
```
kevinschaper, line 59: same as above

glass-ships (approved):

PR #740, opened 2024-06-11
genotype and variant support

kevinschaper:
Adds initial support for genotypes & variants and their associations to the API & UI


amc-corey-cox:
Hey Kevin, so far this looks pretty good to me. Is there any specific thing you would like a review of?

amc-corey-cox (approved):
This looks good to me!
I noticed a bunch of things moved around in the category_enums.py but many of them seem to stay mostly the same. Is there a reason for the rearrangement?

kevinschaper:
The category enums are auto generated, so that might just be random-ness. This PR as a whole has a LOT of autogenerated content, so it's a little tough to spot what's changed. The main thing was that I did need to autogenerate both fixtures and category enums, and I also needed to edit the association_type_group yaml file. I think maybe there's a direction we should go in to just generate those labels most of the time, and only override for things like "Causal Genes" - but we can work on that later.

PR #754, opened 2024-07-06
Fix taxon name colliding with genotype names in search

### LINKED ISSUE #753: Fix results table error that superimposes "Taxon Name" over each row
**Where:** Navigating dev.monarchinitative.org
**When:** 2024/07/05 17h16m
**What:** Using the search bar to find an entity, I typed "B6J.B6N." The table of results was displayed, with the field describing "Taxon Name" superimposed over the results themselves, as shown in the screenshot. Start reading the first entry in the results list, and notice the gray text over the word 'background.' When you hover over that text, the tooltip shows "Taxon Name."

Screen Shot 2024-07-05 at 5 12 50 PM

kevinschaper:
A small css tweak to get long names to wrap so that they don't collide with the taxon label

fixes #753

### DIFF
--- a/frontend/src/pages/explore/TabSearch.vue
+++ b/frontend/src/pages/explore/TabSearch.vue
@@ -78,6 +78,7 @@
           :copy="true"
           color="secondary"
         />
+
       </div>
       <p class="description truncate-3" tabindex="0">
         {{ result.description || "No description available" }}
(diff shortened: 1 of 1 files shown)


kevinschaper:
merging to get it to deployed to beta.mi.org, we can mess with the layout more later, but this solves the collision.

Issue #626, opened 2024-03-05
Ortholog Phenotypes checkbox is only available on a gene that has it's own phenotypes


kevinschaper:
I thought I'd come up with a great solution to allowing Ortholog Phenotypes, showing a checkbox to include ortholog phenotypes as well. 
<img width="863" alt="Screenshot 2024-03-05 at 1 39 40 PM" src="https://github.com/monarch-initiative/monarch-app/assets/4535019/6626f602-5a76-4724-aac2-999d621f91b2">

Unfortunately, there's no way to get to the checkbox if the gene itself doesn't have g2p associations, so we may need to go down the "fake category" route after all.

kevinschaper:
This is actually a much bigger problem for Orthlog Disease, since there's no Disease option in the associations section for non-human animal genes, which means there's not actually a path to the checkbox. 

One solution might be to move the checkbox up above, and have the association-counts re-generated with orthology traversing included, or have that model include both with and without traversing orthology? 

Another solution would be the "fake category" route, where association counts offers counts on something like `biolink:GeneToOrthologToPhenotypicFeature` or maybe `monarch:GeneToOrthologToPhenotypicFeature` ... but I'm still hesitant to implement this feature using anything that looks like a real association category but isn't.

PR #766, opened 2024-07-24
Add direct only option to association table on node pages

### LINKED ISSUE #763: Add controls to show direct associations only on node pages
When investigating non-leaf diseases that have phenotype (or other) associations, it can be difficult to tell if there are direct associations, or to find them if they're present. We should add a some kind of affordance (checkbox? switch?) to limit to direct associations only. 

(Also, it needs to reset the page number if checked!)

kevinschaper:
fixes #763

Adds support for viewing direct associations only using a select box after the association table "sentence" giving the options of "directly" or "including sub-classes", and this select-box only appears if the node actually has subclasses. I think this also sets up a nice pattern to solve #626, since it can show as a select box in the exact same spot, but will show on genes, which won't have subclasses.  

(Not really deploy previewable, since it relies on a backend change as well)


kevinschaper:
ThankyouThankyouThankyou to @amc-corey-cox for helping me work through a way to present this feature that didn't just look like a debugging tool

PR #775, opened 2024-08-05
Show all association tables together

### LINKED ISSUE #734: Update the progressive disclosure on mi.org
No more progressive disclosure. Follow the video Chris made. This refers to:

- Prominently featuring the ‘genes’ associated with the entity used for search

- Getting rid of the drop-down menus in the results tables and, instead, display all results at once, either by emulating the behaviour for results in the “Alliance” [website](https://www.alliancegenome.org/) or something similar to what was implemented before the latest UI update (the column on the left with all possible results, NOT from a drop-down).

We could use HPO API calls where appropriate – when integrating the search (see https://github.com/monarch-initiative/monarch-app/issues/736)

- [ ] show association counts in the title of sections
- [ ] background color / boxes to visually wrap and split sections
- [ ] left justify to the TOC
- [ ] make the scroll spy / TOC visible always
- [ ] remove the idea of expansion and collapse of tables

kevinschaper:
Show all association tables at once on a node page rather than having to select the association type from an option box. From the initial first pass, I think the minimum to make this work is going to also include making the page nav always visible and making sure it has scrollspy functionality. This may turn into a longer lived PR if we find that we need to tweak table content before we want to merge.


amc-corey-cox:
Kevin, I see this still has some linting failures. Would you like me to review it anyway?

amc-corey-cox (approved):
It looks good to me. I didn't extensively test the web function. I also think we should probably present this a little more broadly before we pull the trigger to move it to prod.

amc-corey-cox:
There is a problem with the formatting of section headers when this is merged in. I'm working on finding the source of the issue. Sorry I missed this when I reviewed... Sara pointed it out to me on one of my PR's.

Issue #777, opened 2024-08-06
Update Explore page on mi.org


sagehrke:
Update to this page: https://monarchinitiative.org/explore
- [x] Tools to explore the KG need to be more easily findable. Add links to monarchR, assistanct, API, etc, under the Explore page. 
- [x] Move the graph infographic with node/edge counts that is currently on the home page to the explore page
- [x] Add help icon on the Explore page

sagehrke:
I did a brain dump of my ideas for a future Explore page/navigation only using the buliding blocks we already have. 
Check it out [here](https://docs.google.com/presentation/d/101N-rLMkqcFxOYN0nI3Dc_jIhIPdbbs2KAkmr4uy2cg/edit#slide=id.p). 
This may be out of scope for this ticket, which is all good if so. 
This is also just my ideas and I am not stuck in them. Feel free to take or leave whatever parts make sense to ya'll @kevinschaper @amc-corey-cox @monicacecilia

PR #791, opened 2024-08-16
Monarch organizational home page changes to highlight foundational products

### LINKED ISSUE #672: Configure redirects from monarchinitiative.org
Let's borrow from the convenience of https://github.com/tis-lab/redirects/blob/main/redirects.yaml to create nginx redirects in our nginx config. 

I'm guessing we'll have a step to translate the yaml into nginx redirect format, and then we'll need to use some kind of include directive to bring that into the nginx config.

kevinschaper:
Not my branch (mostly), but we need a PR for it


amc-corey-cox:
This should be ready for review now! Please check it out using the link above in the Netlify Deployment preview above.
@monicacecilia & @sagehrke

amc-corey-cox:
@kevinschaper Can you take  a look at this one? I'd like to get your input before I merge. I think it looks pretty good and I believe we have covered most of the UI redevelopment that we were hoping to get to in this phase.

kevinschaper:
Looks good! I want to make a better destination for the KG link on the home page than just search, but I don't think that needs to be in this PR.

sagehrke (commented):
Thanks for all your work on this @amc-corey-cox! If this PR was only focused on the homepage, please ignore my other critiques. 

## home page
- No updates for now. This is a great first update to the new website interface. 

## explore page
- We are still missing some aspects of this ticket: https://github.com/monarch-initiative/monarch-app/issues/777
- I am also curious what we think of moving the infographic to the top of the page. 

## team page
- See updates here: https://github.com/monarch-initiative/monarch-app/issues/796

PR #800, opened 2024-09-06
Move metadata section on explore page to its own tab

kevinschaper:
Mostly the goal here is to fix the metadata section showing up above the text annotator and phenotype explorer content, but I took it a step further by making it an additional tab.


kevinschaper:
This was merged into another PR and has already come in

PR #808, opened 2024-09-16
Move foundational products above monarch tools, update MonarchR link

kevinschaper:


kevinschaper:
Going to self-approve, since @amc-corey-cox and I worked on this together and reviewed as we went.

PR #890, opened 2024-11-11
Update to eslint.config.js

glass-ships:
eslint has migrated from the `.eslintrc.json` to the new "flat config", `eslint.config.js` style. 

This PR:
- updates to the new config style
- addresses a number of linting/formatting/typing errors that were (for some reason) caught with the new version but not previously
- updates some dependencies and make necessary usage changes
- updates usage of UAParser, as there have been some changes from v1 to v2 (locally tested, should still correctly parse browser/cpu/etc)


glass-ships:
@vincerubinetti Hope you're doing well, long time no see! 
Sorry to drag you back, feel free to ignore. 

But do you have any idea why `axe.test.ts` would be throwing type issues as a result of this PR?  
I've tried digging around a bit but playwright is an..interesting package and it's not super clear to me what, if anything, has changed here.

vincerubinetti:
The issue seems to be with this line:

```ts
type Test = Parameters<typeof test>[1];
```

I assume I wrote it like this originally because Playwright did not conveniently export a (e.g.) `TestFunction` type we could use to define the shape of the `async ({ page, browserName }) => {` function, so I had to extract it manually from the `test()` function like that.

My guess is that in upgrading the version of `@playwright/test`, somewhere along the line they added a function overload to the `test()` function. Now it can be:

```
test("some name", () => {});
// OR
test("some name", { someTestDetail: "abc" }, () => {});
```

which made `Parameters<typeof test>[1];` read the wrong parameter number, i.e. the middle arg of the second overload there, `TestDetails`.

I've found that this works (for now), where these types are just imported from playwright, and this union is defined under-the-hood in `node_modules/playwright/types/test.d.ts`:

```ts
/** test func args */
type TestArgs = PlaywrightTestArgs &
  PlaywrightTestOptions &
  PlaywrightWorkerArgs &
  PlaywrightWorkerOptions;
// it'd be nice if they had a type for this union and exported it for us so we could just use it forever

/** generic page axe test */
const checkPage =
  (path: string, selector?: string) =>
  async ({ page, browserName }: TestArgs) => {
```

Or a better long term solution is to refactor it into this:

```ts
const checkPage = (title: string, path: string, selector?: string) =>
  test(title, async ({ page, browserName }) => {})

checkPage("Accessibility check " + path, path);
```

glass-ships:
Nice, that last refactor seems to have done it!

Thanks so much for taking a look, I'm not sure I'd have ever figured that out haha

glass-ships:
@kevinschaper sorry for the ping - thoughts on this?

amc-corey-cox:
This looks good but I'm not sure if there is something else going on in the app this might impact. @kevinschaper and @ptgolden can one of you take a quick look and make sure this is okay to bring in. It looks alright to me.

PR #892, opened 2024-11-14
Set a unique key on phenogrid cells and use it as key when rendering

### LINKED ISSUE #755: Fix Phenogrid results error when changing order (sort) of input phenotypes
Where: dev.monarchinitative.org
When: 2024/07/07 21h21m
What: When inspecting the results of Phenotype Explorer, I changed the order of the results using the 'Sort' dropdown menu. The order changed, but the information about the comparison and values for each score still corresponded to the previous pair. 

To reproduce: 
I entered these phenotypes: MP:0008752,MP:0008750,MP:0008751,MP:0004045,MP:0010170,MP:0020449,MP:0003691,MP:0003884,MP:0001599,MP:0011986,MP:0008596,MP:0008553,MP:0011080,MP:0002078,MP:0001798,MP:0008735
To compare them to "phenotypes from these genes/diseases..." (selected from dropdown menu)| "Human Diseases" (selected from dropdown menu), using AIC (selected from dropdown menu).

The resulting Phenogrid shows this information when I hover over the upper left square in the grid (screenshot). 
Screen Shot 2024-07-07 at 10 18 55 PM

When I use the 'Sort' option to reorganize by highest score, the phenotypes are reorganized, but when I hover over the same square, the same information that appears belongs to the previously compared phenotypes. That information is NOT being reorganized (screenshot). This is an error. 

Screen Shot 2024-07-07 at 10 20 10 PM

Screen Shot 2024-07-07 at 10 19 06 PM

ptgolden:
Previously, the key on the element being rendered was the index of the object in the array of cells. When cells were reordered (or, I imagine, transposed), they were not being re-rendered because even though the array was shuffled, the index for each rendered cell stayed constant.

This commit creates a compound identifier for each cell consisting of the ID of the row and the ID of the column. This uniquely identifies each cell in the grid.

Fixes #755

### DIFF
--- a/frontend/src/components/ThePhenogrid.vue
+++ b/frontend/src/components/ThePhenogrid.vue
@@ -155,8 +155,8 @@
         <!-- cells -->
         <g class="cells">
           <tooltip
-            v-for="(cell, index) in cells"
-            :key="index"
+            v-for="cell in cells"
+            :key="cell.id"
             :interactive="true"
             placement="bottom"
             :append-to="appendToBody"
@@ -454,6 +454,7 @@ const cells = computed(() =>
     )
     .flat()
     .map(({ col, row }) => ({
+      id: `${col.id}|${row.id}`,
       col,
       row,
       ...getCell(col, row),


ptgolden:
The only way this key wouldn't uniquely identify a cell is if there are repeated elements in the row/column, but repeats there seem to be eliminated. (e.g. <https://monarchinitiative.org/phenogrid-search?subjects=HP:0000939,HP:0000939&object-group=Human+Diseases>)

kevinschaper (approved):

Issue #440, opened 2023-10-24
Make histopheno bars clickable to filter association table


kevinschaper:
Right now we use an [in-code slim](https://github.com/monarch-initiative/monarch-app/blob/5df6275ff2fbc75ebc1dc4108401747936c3cbf6/backend/src/monarch_py/datamodels/solr.py#L14) to generate the histopheno graph, which serves as a non-interactive summary of disease to phenotype associations. 

Keeping the visualization the same, we could also summarize other associations by creating or borrowing similar slims. 

We could also consider including a ribbon with the associations section that would fill the same role, as well as being navigational for subsetting.

caufieldjh:
Are there plans to make the resulting data (as per the slim, since it may not be obvious to users what slim is being used) or graph downloadable?

kevinschaper:
> Are there plans to make the resulting data (as per the slim, since it may not be obvious to users what slim is being used) or graph downloadable?

Right now it's possible to download an image or a table with counts from the histopheno section, but I think that clicking on a bar in the chart really should give you a table of those annotations, and then that table should be downloadable

kevinschaper:
Hi @pkalita-lbl, before I go too far down the road of awkwardly cobbling together two neighboring UI elements, could you spend some time making a mock up of what we could do for a Phenotypes section that combines some kind of visualization of phenotypes with the detailed results?  I think probably following the well worn path of a ribbon plus table probably makes sense, but it seems like we have a slim that's slightly smaller than other ribbon scenarios, so maybe that leaves room for other approaches?

sagehrke:
@pkalita-lbl Can you please link the mockups for the phenotype section for others to review? Thank you!

pkalita-lbl:
Ah yes sorry about that. I sent these directly to Kevin and neglected to update the issue. Here are two sketches that illustrate:

* Keeping the bar chart: https://design.penpot.app/#/view/75aa5258-9a82-8002-8003-88ba6a2daebd?page-id=75aa5258-9a82-8002-8003-88ba6a2daebe&section=interactions&index=0&share-id=af8aaf7c-05e6-8124-8003-8cc4459fc3b9
* Using a ribbon: https://design.penpot.app/#/view/75aa5258-9a82-8002-8003-88ba6a2daebd?page-id=eb151800-053e-8018-8003-88c9f4733b8f&section=interactions&index=0&share-id=75aa5258-9a82-8002-8003-8cc4d9041914

[...]

Issue #430, opened 2023-10-23
Add experimental edge-merged association table endpoint


kevinschaper:
It's clear that a lot of people still have the expectation that association tables will have edged merged on some natural key (I think subject+negation+predicate+object). We made the architectural choice that we wanted to avoid a merge step on the graph (plus the evidence graph to explain it). To see if we can stick with that choice we should attempt to use solr result grouping to mimic that behavior. 

- [x] Add a single valued string field to the solr schema concatenating subject, predicate, object and negation
- [ ] Add a grouped association table endpoint that groups on the new field

We might be able to use facet pivoting to get info across the group, otherwise I think info about the group would need to be pulled from the first record of each group.

kevinschaper:
Here's an example of a grouped response that collapses the duplicate-ish association from #328 

```
http://localhost:8983/solr/association/select?q=*:*&group=true&group.limit=5&group.field=grouping_key&fq=subject:HGNC\:555&fq=object:HP\:0000718&fl=subject,predicate,object


{
  "responseHeader": {
    "status": 0,
    "QTime": 1,
    "params": {
      "q": "*:*",
      "group.limit": "5",
      "fl": "subject,predicate,object",
      "fq": [
        "subject:HGNC\:555",
        "object:HP\:0000718"
      ],
      "group.field": "grouping_key",
      "group": "true"
    }
  },
  "grouped": {
    "grouping_key": {
      "matches": 2,
      "groups": [
        {
          "groupValue": "HGNC:555🍪🍪biolink:has_phenotype🍪HP:0000718",
          "doclist": {
            "numFound": 2,
            "start": 0,
            "numFoundExact": true,
            "docs": [
              {
                "predicate": "biolink:has_phenotype",
                "subject": "HGNC:555",
                "object": "HP:0000718"
              },
              {
                "predicate": "biolink:has_phenotype",
                "subject": "HGNC:555",
                "object": "HP:0000718"
              }
            ]
          }
        }
      ]
    }
  }
}
```

kevinschaper:
This one was a stretch goal for November, and may continue to be low priority until we explicitly bring in an ingest that creates a large number of associations that would come up as "duplicates" from different sources. This first step looks promising, next step will be actually making the API endpoint. 

It would probably have been better to make the Solr support and the API endpoint into different issues, but for now I'll just move this issue forward to another release. I think I'll skip past December and mark this one for January.  (where it still may be a lower priority, given holidays etc)

[...]

Issue #841, opened 2021-05-26
Add an option to show all phenotypes


jmcmurry:
Pagination is good but it would be better to be able to see all at once, preferably grouped by area as on hpo.jax.org

https://monarchinitiative.org/disease/MONDO:0007947#phenotype

sagehrke:
Should this ticket stay open? Thanks!

Issue #821, opened 2019-07-02
In results table, group relations per gene, avoid single entry per relation.


monicacecilia:
On results pages, @mellybelly wants to see a block that displays information per gene - that is, all relations per gene, and not a single entry for each relation a gene has with each of the associated diseases. 

For example, in https://beta.monarchinitiative.org/disease/OMIMPS:PS214100#gene, 28 of the 31 associations are '*causes condition*;' Melissa would like to see these grouped and displayed in a block.

![Screen Shot 2019-07-01 at 7 21 25 PM](https://user-images.githubusercontent.com/1714621/60477912-a364c200-9c35-11e9-9569-cb2dcc2d558e.png)

monicacecilia:
Happy Birthday, dear ticket. 🍰 🍰 🍰 🍰 🍰
Moving to app repo. Need to revisit.

sagehrke:
@kevinschaper For some reason I thought we fixed this. But did we actually fix it?

PR #914, opened 2024-11-26
Add support for Solr group queries to Solr service

ptgolden:
This adds a new method for the Solr service, `.group_query()`, which allows running Solr queries which group returned documents.

<https://solr.apache.org/guide/6_6/result-grouping.html>

This only implements the `group.query` option, not `group.field` or `group.func`.

### Related issues

- Necessary for #430 and #440

### Checks

- [x] All tests have passed (or issues created for failing tests)


ptgolden:
(Also related to #821 and possibly #841)

ptgolden:
It's possible that `grouped_query` might be a better name for the method-- what do you think @kevinschaper?

PR #930, opened 2024-12-13
Update orthology association table

kevinschaper:
A first try at largely keeping our existing association table infrastructure but defining each section separately, doing the ortholog table specifically.

### DIFF
--- a/backend/src/monarch_py/solr_cli.py
+++ b/backend/src/monarch_py/solr_cli.py
@@ -210,11 +210,7 @@ def multi_entity_associations(
     ] = None,
     counterpart_category: Annotated[
         Optional[List[str]],
-        typer.Option(
-            "--counterpart-category",
-            "-c",
-            help="A comma-separated list of counterpart categories"
-        ),
+        typer.Option("--counterpart-category", "-c", help="A comma-separated list of counterpart categories"),
     ] = None,
     limit: fields.LimitOption = 20,
     offset: fields.OffsetOption = 0,
@@ -306,9 +302,7 @@ def histopheno(
 
 @solr_app.command("association-counts")
 def association_counts(
-    entity_id: Annotated[
-        str, typer.Argument(help="The entity to get association counts for")
-    ],
+    entity_id: Annotated[str, typer.Argument(help="The entity to get association counts for")],
     fmt: fields.FormatOption = fields.OutputFormat.json,
     output: fields.OutputOption = None,
 ):
--- a/frontend/src/components/AppTable.vue
+++ b/frontend/src/components/AppTable.vue
@@ -454,21 +454,6 @@ const ariaSort = computed(() => {
   text-transform: capitalize;
 }
 
-/* first th */
-.th:nth-child(1) {
-  width: 30%;
-}
-
-/* second th */
-.th:nth-child(2) {
-  width: 5%;
-}
-
-/* third th */
-.th:nth-child(3) {
-  width: 30%;
-}
-
 /** body cells */
 .td {
   border-bottom: solid 2px $light-gray;
--- a/frontend/src/pages/node/AssociationsTable.vue
+++ b/frontend/src/pages/node/AssociationsTable.vue
@@ -222,7 +222,6 @@ const orthologColoumns = computed<Cols<Datum>>(() => {
       slot: "taxon",
       key: "taxon" as Datum,
       heading: "Taxon",
-      sortable: true,
     },
     {
       slot: "ortholog",
@@ -238,7 +237,6 @@ const orthologColoumns = computed<Cols<Datum>>(() => {
       slot: "primary_knowledge_source",
       key: "primary_knowledge_source" as Datum,
       heading: "Source",
-      sortable: true,
     },
     { slot: "divider" },
     {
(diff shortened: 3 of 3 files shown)


backend/src/monarch_py/solr_cli.py
```diff
@@ -210,11 +210,7 @@ def multi_entity_associations(
     ] = None,
     counterpart_category: Annotated[
         Optional[List[str]],
-        typer.Option(
-            "--counterpart-category",
-            "-c",
-            help="A comma-separated list of counterpart categories"
-        ),
+        typer.Option("--counterpart-category", "-c", help="A comma-separated list of counterpart categories"),
```
kevinschaper, line 213: This was autoformatting from `make format`

backend/src/monarch_py/solr_cli.py
```diff
@@ -306,9 +302,7 @@ def histopheno(
 
 @solr_app.command("association-counts")
 def association_counts(
-    entity_id: Annotated[
-        str, typer.Argument(help="The entity to get association counts for")
-    ],
+    entity_id: Annotated[str, typer.Argument(help="The entity to get association counts for")],
     fmt: fields.FormatOption = fields.OutputFormat.json,
     output: fields.OutputOption = None,
 ):
```
kevinschaper, line 308: autoformatting

frontend/src/components/AppTable.vue
```diff
@@ -454,21 +454,6 @@ const ariaSort = computed(() => {
...
-.th:nth-child(2) {
-  width: 5%;
-}
-
-/* third th */
-.th:nth-child(3) {
-  width: 30%;
-}
-
 /** body cells */
```
kevinschaper, line 457: This is where I might be causing trouble by altering potentially broader impact css, but I _think_ I'm actually reducing the trouble that this overly broad css was causing. This PR is the first of many for defining the columns for each association type separately, and it doesn't make sense that we'll always want these widths. Plus, if you scan through Human Disease on this branch, most tables look better, in particular it gives as much space as possible to disease models.

frontend/src/pages/node/AssociationsTable.vue
```diff
@@ -24,6 +24,31 @@
...
+          id: row.subject,
+          name: row.subject_label,
+          category: 'biolink:Gene',
+          info: row.subject_taxon_label,
+        }"
+        :breadcrumbs="getBreadcrumbs(node, row, 'object')"
+      />
+    </template>
+
     <!-- subject -->
```
kevinschaper, line 52: how to display an ortholog cell

frontend/src/pages/node/AssociationsTable.vue
```diff
@@ -65,6 +90,15 @@
...
 
+    <template #has_evidence="{ row }">
+      <AppLink
+        v-for="(source, index) in row.has_evidence_links"
+        :key="index"
+        :to="source.url || ''"
+        >{{ source.id }}</AppLink
+      >
+    </template>
+
```
kevinschaper, line 101: how to display a has_evidence cell (borrowed from SectionAssociationDetails)

frontend/src/pages/node/AssociationsTable.vue
```diff
@@ -180,8 +214,46 @@ const search = ref("");
...
+    { slot: "divider" },
+    {
+      slot: "details",
+      key: "evidence_count",
+      heading: "Details",
+      align: "center",
+    },
+  ];
+});
+
```
kevinschaper, line 250: which columns should be shown for the ortholog section

frontend/src/components/AppTable.vue
amc-corey-cox: Yeah, we really need the new UI person to do a CSS review across the app but this looks good for now.

frontend/src/pages/node/AssociationsTable.vue
amc-corey-cox: I really like how you are using this data structure.

amc-corey-cox (approved):
This looks good to me. I added some inline comments but just to tell you I think we need an overall review of the CSS and that I liked the ortholog columns data structure.

frontend/src/pages/node/AssociationsTable.vue
ptgolden: You likely shouldn't be using `index` as a key here. If the table gets resorted, this may not be re-rendered. See the [Vue documentation](https://vuejs.org/api/built-in-special-attributes#key), and also #892 for a fix to a similar bug.

The key should uniquely identify what objcet this is rendering. I'm not totally familiar with this data structure, but I'm guessing that would be `source.id`.

For example, if I had the set of rows:
```
[
  { source: { id: 'a' }}, // index is 0
  { source: { id: 'b' }}, // index is 1
  { source: { id: 'c' }}, // index is 2
]
```

Vue would render if fine initially. But if I resorted those rows:

```
[
  { source: { id: 'c' }}, // index is 0
  { source: { id: 'b' }}, // index is 1
  { source: { id: 'a' }}, // index is 2
]
```

Vue would not know that it would need to re-render those rows, since the key of the elements never changes-- the first element always has index 0, the second index 1, and so on.

kevinschaper: fixed! thanks for catching that!

PR #937, opened 2025-01-09
Add medical action association table to UI (plus fixture & enum regen)

### LINKED ISSUE #919: Update UI for MAxO
There should be a Medical Actions table on disease, and phenotype pages. Disease and Phenotype tables should also show up on Medical Action pages. 

This probably needs at least some API side changes, so that the ChemicalOrDrugOrTreatmentToDiseaseOrPhenotypeAssociations (or something like that) are labeled nicely. 

An addition to https://github.com/monarch-initiative/monarch-app/blob/main/backend/src/monarch_py/utils/association_type_mappings.yaml might be enough to both opt these associations into being displayed on the page, and set the label. 

On the UI side, assuming that we'll use the standard association table, and just need to customize columns.

### LINKED ISSUE #935: Pytest seems to be failing in certain environments
I've noticed some CI pipelines for test-backend passing, but locally as well as for other PR's (like #890 and #745) pytest is failing with something along the lines of:

```
ImportError: Error importing plugin "tests.fixtures.search_query": No module named 'tests'
```

Appears to be an issue with pytest trying to run/import `conftest.py` but I can't tell when this issue started or why it's complaining now (pytest version doesn't appear to have changed, and pinning to 8.2.0 doesn't fix the issue)

kevinschaper:
Adds a medical action table showing the MaXO term, predicate, phenotype and disease. For simplicity sake, the section is labeled as "Medical Action" on both MaXO pages and phenotype pages. 

Creating as a draft, because I realized that it needs to show up on disease pages as well, which means that the API request will need to change so that when "entity=<ID of page>" comes into the api request, disease_context_qualifier is searched, and I think we'll also want to make sure we're searching disease_context_qualifier_closure as well (which is hopefully already getting generated!

Closes #919 
Closes #935


backend/src/monarch_py/implementations/spacy/spacy_implementation.py
```diff
@@ -32,22 +33,31 @@ def init_spacy(self, grounding_implementation: GroundingInterface):
...
+                    if d.is_dir() and d.name.startswith("en_core_sci") and "egg-info" not in d.name
+                ),
+                None,
+            )
             if inner_model_dir:
                 # Load the model
-                self.nlp = spacy.load(str(str(model_archive.parent / model_subdir.name / inner_model_dir.name / model_subdir.name)))
+                self.nlp = spacy.load(
+                    str(str(model_archive.parent / model_subdir.name / inner_model_dir.name / model_subdir.name))
+                )
```
kevinschaper, line 54: Everything in this file is just formatting changes that got picked up

.github/workflows/test-backend.yaml
```diff
@@ -4,6 +4,9 @@ on:
   pull_request:
   workflow_dispatch:
 
+env:
+  PYTHONPATH: ${{ github.workspace }}/backend
+
```
kevinschaper, line 9: fix for https://github.com/monarch-initiative/monarch-app/issues/935

kevinschaper:
@amc-corey-cox  @ptgolden The more straightforward part of this is the specific handling for MaXO associations in the UI. The weird part is that by default, we want disease pages to return these MaXO associations in the association-counts part of the entity response, and be able to fetch an association table for MaXO associations where the ID of the page that we're on is a MONDO ID which is in the disease_context_qualifier or disease_context_qualifier_closure field. I handled that by expanding what is matched in the API endpoint for the `entity` param, since we intend that to return associations involving that entity, and this is the first situation we have where we want the association to show up on a page based on the ID appearing in a qualifier field. I don't know if it's a 100% perfect solution, but I talked myself into the idea that it's a pretty reasonable one.

backend/src/monarch_py/datamodels/category_enums.py
```diff
@@ -4,119 +4,60 @@
...
-    SI_RNA = "biolink:SiRNA"
-    SNV = "biolink:Snv"
-    STUDY = "biolink:Study"
-    STUDY_VARIABLE = "biolink:StudyVariable"
-    TRANSCRIPTION_FACTOR_BINDING_SITE = "biolink:TranscriptionFactorBindingSite"
-    TREATMENT = "biolink:Treatment"
-    WEB_PAGE = "biolink:WebPage"
-    ZYGOSITY = "biolink:Zygosity"
+    MOLECULAR_ENTITY = "biolink:MolecularEntity"
+    LIFE_STAGE = "biolink:LifeStage"
```
amc-corey-cox, line 23: I'm going to presume that all of these deletions and additions make sense, although I admit I didn't check to make sure we don't need these.

backend/src/monarch_py/datamodels/category_enums.py
```diff
@@ -4,119 +4,60 @@
...
         "biolink:DiseaseOrPhenotypicFeatureToGeneticInheritanceAssociation"
     )
-    VARIANT_TO_DISEASE_ASSOCIATION = "biolink:VariantToDiseaseAssociation"
     CAUSAL_GENE_TO_DISEASE_ASSOCIATION = "biolink:CausalGeneToDiseaseAssociation"
-    CHEMICAL_TO_DISEASE_OR_PHENOTYPIC_FEATURE_ASSOCIATION = "biolink:ChemicalToDiseaseOrPhenotypicFeatureAssociation"
+    DISEASE_OR_PHENOTYPIC_FEATURE_TO_LOCATION_ASSOCIATION = "biolink:DiseaseOrPhenotypicFeatureToLocationAssociation"
+    CHEMICAL_OR_DRUG_OR_TREATMENT_TO_DISEASE_OR_PHENOTYPIC_FEATURE_ASSOCIATION = (
+        "biolink:ChemicalOrDrugOrTreatmentToDiseaseOrPhenotypicFeatureAssociation"
+    )
+    VARIANT_TO_PHENOTYPIC_FEATURE_ASSOCIATION = "biolink:VariantToPhenotypicFeatureAssociation"
```
amc-corey-cox, line 60: Again, I'll presume your reorganization is for improving readability or logical consistency.

backend/src/monarch_py/implementations/solr/solr_parsers.py
```diff
@@ -258,6 +258,13 @@ def get_association_direction(entity: List[str], document: Dict) -> AssociationD
         document.get("object_closure") and any(e in document.get("object_closure") for e in entity)
     ):
         direction = AssociationDirectionEnum.incoming
+    elif document.get("disease_context_qualifier") in entity or (
+        document.get("disease_context_qualifier_closure")
+        and any(e in document.get("disease_context_qualifier_closure") for e in entity)
+    ):
+        # This is a special case for disease_context_qualifier, if an association between two other entities
+        # only occurs within the context of a disease, we can treat it like an incoming association
+        direction = AssociationDirectionEnum.incoming
```
amc-corey-cox, line 267: Codecov is complaining that 261 and 267 aren't covered by tests. It seems reasonable to ignore this but if it's easy to make a test here it would be nice to cover these edge cases.

backend/src/monarch_py/implementations/spacy/spacy_implementation.py
amc-corey-cox: Check, I'm ignoring it then.

amc-corey-cox (approved):
I didn't review the big diffs in tests/ or frontend/fixtures but I don't see anything that jumps out as concerning to me. I flagged a couple of things just to make sure you noticed them and that they were done on purpose.
There doesn't appear to be a deploy preview. Would you like me to review the frontend view?

PR #943, opened 2025-01-21
Additional phenogrid fixes

ptgolden:
Allow `<b>` tags in node labels (including phenogrid labels), and add appropriate keys for phenogrid row/column tooltips.

### DIFF
--- a/frontend/src/components/AppNodeText.vue
+++ b/frontend/src/components/AppNodeText.vue
@@ -45,7 +45,7 @@ const props = withDefaults(defineProps<Props>(), {
 
 const container = ref<HTMLSpanElement | SVGTSpanElement | null>(null);
 
-type ReplacedTag = "sup" | "a" | "i";
+type ReplacedTag = "sup" | "a" | "i" | "b";
 
 type Replacement = {
   type: ReplacedTag;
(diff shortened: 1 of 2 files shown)


amc-corey-cox (approved):
Patrick walked me through these. Looks good.

dpavam:
Hi @ptgolden, thanks for pushing these changes! 
I was trying these on our website. The tooltips now transpose correctly, thanks!  

As for the `<b>` tags, I have tried incorporating them on a local version of the website:
By passing `<b>label</b>` I get the following behaviour:

Screenshot 2025-01-22 at 12 12 08

The label inside the tooltip is in bold 👍  but the label on the y-axis prints the `<b>` tags instead of going bold. 
Am I missing something else that I should pass to the phenogrid?

Not sure if this is something on my side, but I thought I should let you know before the release this week.

ptgolden:
Oh shoot- let me fix that.

ptgolden:
@dpavam this was a little more involved than I thought, but the fix is in the PR I just made

dpavam:
Thanks @ptgolden! Has this been released yet? I have tested this with the current main branch after the PR but the `<b>` tag literals are still showing on the labels. Is there any chance you have the example case/implementation you used to check this was working?

PR #944, opened 2025-01-21
When serializing nodes for links, only include small subset of data

### LINKED ISSUE #912: Entities with many subclasses load super slowly
example: http://dev.monarchinitiative.org/PATO:0000001

I'm not sure if this is a backend problem (something about the hierarchy code is silly, like making one request per subclass entity), or if it's a rendering problem (something in the AppNodeBadge is expensive?). 

If there's lots of associations, but few subclasses, everything seems fine, so it does seem to be specific to the hierarchy API request or UI components.

ptgolden:
Previously, if an AppLink was rendered from a node, the entire node was serialized. Since a node is a non-trivially large JSON structure, this resulted in a lot of wasted serialization, as only a few pieces of information are necessary to generate a link. As a result, pages that rendered a non-trival amount of links (in this case, SectionHierarchy) would be very laggy as the browser struggled to do all that serialization.

This commit filters only those properties necessary to render a link: `id`, `category`, `name`, and `in_taxon_label`.

(Note that this stretegy is already used in AssociationSummary.vue).

Fixes #912

### DIFF
--- a/frontend/src/global/breadcrumbs.ts
+++ b/frontend/src/global/breadcrumbs.ts
@@ -3,15 +3,32 @@ import { pick } from "lodash";
 import type { DirectionalAssociation, Node } from "@/api/model";
 import { parse } from "@/util/object";
 
-const breadcrumbKeys = ["id", "category", "name", "in_taxon_label"] as const;
+/**
+ * This constant represents the minimal set of properties needed to render a
+ * link to a Node object. If more properties must be added to render such a link
+ * in the future, add them to this list.
+ *
+ * This is necessary because some pages render a large number of links to nodes.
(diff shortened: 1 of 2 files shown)


amc-corey-cox (approved):
Patrick will add a description of what toBreadcrumbNode does and why we need it to the function.

ptgolden:
The same function should be used in `pages/node/AssociationsTable.vue` and `pages/node/AssociationsSummary.vue`

ptgolden:
> The same function should be used in `pages/node/AssociationsTable.vue` and `pages/node/AssociationsSummary.vue`

I did not end up doing this, because a DirectionalAssociation is different than a Node.

PR #945, opened 2025-01-30
Allow truncating SVG in AppNodeText

ptgolden:
Labels in the phenogrid were previously rendered as `<text>` SVG elements, rather than `<AppNodeText>` Vue elements which whitelist certain markup. This replaces those `text` elements with `AppNodeText` elements.

An issue, though, is that labels were previously truncated based on their bare text length. So the string `'label<b>with bolded</b>'` might be truncated to `'label<b'` or `'label<b>with bo'` rather than `'label with bo'` (where "with bo" would be bolded).

This PR fixes that bug by truncating text according to its length as it appears in rendered SVG.

Fixes the bug mentioned in https://github.com/monarch-initiative/monarch-app/pull/943#issuecomment-2607114471.


ptgolden:
I am going to add several test cases before continuing

amc-corey-cox (approved):
This is a great change! This makes it much simpler and we don't have to rely as much on our own internal logic. This both solves the bug and makes the whole thing easier to understand and maintain.

amc-corey-cox:
Waiting to merge until you make that last small fix.

sagehrke:
@ptgolden this is going to have conflicts now with all the updates @varun-divya has done. If it still needs to happen, please work with her to get the conflicts resolved. Thanks!

ptgolden:
> 1. Security Concern - Potential XSS Vulnerability (Medium Priority)

I'm aware of this concern and it was introduced elsewhere. I'm going to make a new issue about it, but it will take some work to fix that is outside the scope of this PR.

> Using <foreignObject> for SVG rendering (line 12) is clever but has limitations:

Unfortunately, reliably rendering superscripted text with SVG has inescapable cross-browser inconsistencies. See my commit message in 48457da26008c207f870ec6a8b06a1f692ae9b35. This approach should be much better browser-compatibility wise.

 I take the accessbility comment, but when this component is rendered in SVG, it's going to be in the case of a data visualization component, which would not be fit for a screen reader anyway.

PR #953, opened 2025-02-14
Bugfix/issue -744 - Page is scrolling while entering text input or sorting data

varun-divya:
### Checks


kevinschaper:
This one is extra challenging! If there's just one row in the table, it winds up kind of strange.

Screenshot 2025-02-17 at 4 30 00 PM

It's would be a bigger change, but it's probably worth exploring having the search bar above the table, rather than fighting with the table size.

kevinschaper (approved):

PR #963, opened 2025-02-25
Add documentation to script for updating publications

ptgolden:
Add documentation to the script to update publications from Google Scholar. Also fix a bug where the update script wouldn't work if the cached data did not yet exist.


amc-corey-cox (approved):
This looks good. Thanks for updating all of the doc strings. The logic makes sense and I like the approach.

PR #964, opened 2025-02-25
Make the makefile work with Poetry v2 as well as v2

### LINKED ISSUE #962: Makefile incompatible with Poetry 2
In Poetry 2, the `-C` flag actually changes directories rather than just using the specified directory as the location of the Poetry environment. Some commands in our Makefile depend on not changing directories.

The options for fixing this are:

1. Use the new `--project` flag when Poetry 2 is detected (and use `-C` for Poetry 1)
2. Still use `-C` but use absolute paths for directories in CLI commands rather than depending on being in the root directory
3. Remove usage of `-C`, prepend any poetry invocation with `cd backend &&`, and assume `$CWD` will always be `monarch_app/backend`

ptgolden:
Closes #962

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-RUN = poetry -C backend run
+RUN = cd backend && poetry run
 VERSION = $(shell poetry -C backend version -s)
 ROOTDIR = $(shell pwd)
 SCHEMADIR = $(ROOTDIR)/backend/src/monarch_py/datamodels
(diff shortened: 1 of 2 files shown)


amc-corey-cox:
#966 is a duplicate.

amc-corey-cox:
@kevinschaper this is causing docs not to build now for the PR I just merged so we should try to get this in soon.

kevinschaper (approved):

PR #979, opened 2025-03-11
Bump action versions

glass-ships:
Combines existing dependabot PR's for bumping action versions which show failing.

Based off latest main to see if tests are actually failing.

Appears to be all good!


amc-corey-cox (approved):
Yep this looks great! Thanks for bumping this for us @glass-ships always nice to see your work!

PR #1002, opened 2025-03-17
Issue 978 - Highlight Search Text in Association Table

### LINKED ISSUE #978: Add matching text to association search
When filtering results with the search box, it matches against more fields than we actually display. We should include matching text in the API response, and find a way to display it in the table UI.

varun-divya:
### Related issues

- Closes #978 

### Summary

- adds ...
- changes ...
- fixes ...
- removes ...

### Checks


amc-corey-cox (approved):
This looks really nice @varun-divya . Great Work! I'll give @kevinschaper a bit to weigh in and then I'll merge it.

PR #1032, opened 2025-04-01
Use `ruff format` for code formatting

### LINKED ISSUE #96: Offload accessibility and positioning burden of select components to 3rd party libraries
Right now, I'm using vue-tippy for tooltip position, accessibility, and styling. But for multi select components, which also have dropdowns/popovers, I'm doing all of that manually. With positioning, I'm doing it very simply, with no sliding/flipping/etc when the dropdown goes beyond the screen.

Two main libraries to look at:

- the brand new and cleanly refactored [floating ui](https://floating-ui.com/docs/react-dom-interactions). used for correctly positioning tooltips/popovers/menus/etc. might also be usable for accessibility in the future, if they ever make a Vue version of their [`use` React hooks](https://floating-ui.com/docs/useListNavigation). might also be used to replace my custom modal code as well.
- [vue combo blocks](https://github.com/sssmi/vue-combo-blocks), a Vue version of the React library [downshift](https://www.downshift-js.com/use-combobox) which has a very convenient and clean way to get all the accessibility props you need for a select component. it doesn't have many users at the moment, but there must be an alternative. and as stated above, hopefully floating ui will implement this soon, so we only need one library for all of this.

### LINKED ISSUE #965: Use ruff as code formatter
Ruff's formatter is a ["drop-in replacement for Black"](https://docs.astral.sh/ruff/formatter/). Might as well remove our `black` dependency.

ptgolden:
### Related issues

- Closes #965

### Summary

Drops `black` as a dependency, in favor of using `ruff format` as a formatter. On my machine, it's a ~100x speedup.

There's a small change to the makefile too. Instead of using the `--exit-zero` flag in `ruff check`, the command is prefixed with `-` to [continue running even after a non-zero exit code](https://www.gnu.org/software/make/manual/make.html#Errors-in-Recipes).

### Checks

- [x] All tests have passed (or issues created for failing tests)


ptgolden:
I'm not sure why the frontend linting test is failing. Is it not passing on `main`?

amc-corey-cox:
Hmm, @ptgolden I tried re-running the lint test again and it still failed. When you get a chance can you take another look at this?

glass-ships:
~~I tried running `bun run lint` locally to see if i could just patch this real quick, but it's hanging on some new errors I haven't gotten before...~~

Depending on how we feel about #890, that has a number of frontend dependency updates and makes use of now-deprecated methods of eslint config etc. that may resolve this. 

Edit: I was apparently missing some dependencies, `bun install` got me back in working order.  
It may also be worth considering [pre-commit](https://pre-commit.com) and [pre-commit ci](https://pre-commit.ci), we use it pretty extensively at ORNL and it's been really nice!

glass-ships:
rebased from main to include the switch to `uv` from poetry

glass-ships (approved):

PR #1085, opened 2025-05-23
Fix off by one error on association tables

### LINKED ISSUE #880: Fix off by one error on association tables
In #849, it stated "If there are exactly 5 associations, it allows me to advance to next to show an empty list in the range "6-5". Off by one error?". 

This needs to be fixed.

ptgolden:
Fixes #880.

In an association table, if there are `n` items to display; a page limit of `limit`; and `mod(n, limit)` is 0, then the "next page" and "last page" buttons are not disabled on what should be the last page. This fixes that!

### DIFF
--- a/frontend/src/components/TheTableContols.vue
+++ b/frontend/src/components/TheTableContols.vue
@@ -46,14 +46,14 @@
       <template v-if="showControls">
         <AppButton
           v-tooltip="'Go to next page'"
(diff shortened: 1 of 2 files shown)


amc-corey-cox (approved):
Great work Patrick. Thanks for putting in the extra tests and verifying this.

PR #1103, opened 2025-06-13
Switch from poetry to uv

glass-ships:
Updates pyproject.toml, documentation, makefile, and github actions to use `uv` instead of `poetry`. 

Dependency resolution is much faster since uv is written in Rust, and all poetry functionality we currently use is also available in uv. For instance, in a fresh environment, uv installs monarch-py and all its dependencies in less than 1 minutes, where poetry has historically taken as long as an hour depending on a variety of circumstances. 

One minor thing to note is that, similar to the poetry configuration, `version` in the pyproject.toml is set to 0.0.0, getting automatically set by the github action to be whatever the latest tag number is. 

There does exist an equivalent [uv-dynamic-versioning](https://pypi.org/project/uv-dynamic-versioning/) and a [git issue with uv](https://github.com/astral-sh/uv/issues/3957) to track implementation of being able to set `dynamic = ["version"]` in the pyproject.toml. 

However, this is just a nice to have, and shouldn't affect any workflow or git actions.  
The locally installed version will just show up as 0.0.0, which also currently happens with poetry.


backend/tests/unit/test_curie_service.py
```diff
@@ -3,32 +3,32 @@
...
+        ("HGNC:5", "identifiers.org/hgnc/5"),
         ### These prefixes are currently broken in the curie service
-        ("ENSEMBL:ENSG00000157764", "http://identifiers.org/ensembl/ENSG00000157764"),
-        ("OMIM:613647", "http://identifiers.org/mim/613647"),
-        ("UBERON:0000001", "http://purl.obolibrary.org/obo/UBERON_0000001"),
-        ("WormBase:WBGene00000001", "http://identifiers.org/wb/WBGene00000001"),
+        ("ENSEMBL:ENSG00000157764", "identifiers.org/ensembl/ENSG00000157764"),
+        ("OMIM:613647", "identifiers.org/mim/613647"),
+        ("UBERON:0000001", "purl.obolibrary.org/obo/UBERON_0000001"),
+        ("WormBase:WBGene00000001", "identifiers.org/wb/WBGene00000001"),
```
glass-ships, line 28: At some point, some curie expansions started giving http or https where before they were giving the opposite - I'm not sure what resulted in this difference. 
These changes just give the test a little more wiggle room while asserting that the important content of the URI itself is correct

backend/src/monarch_py/interfaces/grounding_interface.py
```diff
@@ -1,12 +1,13 @@
...
+from typing import List
 
 from monarch_py.datamodels.model import Entity
 
 
 class GroundingInterface(ABC):
     """Abstract interface for grounding text to entities"""
 
-    def ground_entity(self, text: str) -> Entity:
+    def ground_entity(self, text: str) -> List[Entity]:
```
glass-ships, line 10: Doesn't affect runtime, but is a useful fix for IDE linters/type checkers

backend/src/monarch_py/implementations/spacy/spacy_implementation.py
```diff
@@ -15,8 +15,8 @@
 class SpacyImplementation(TextAnnotatorInterface):
     """Implementation of Monarch Interfaces for SPACY"""
 
-    nlp = None
-    grounding_implementation = None
+    nlp: Optional[spacy.language.Language] = None
+    grounding_implementation: Optional[GroundingInterface] = None
```
glass-ships, line 19: Same thing here - just helps type checking

backend/src/monarch_py/api/main.py
```diff
@@ -16,11 +17,12 @@
...
 
 
-@app.on_event("startup")
-async def initialize_app():
+@asynccontextmanager
+async def lifespan(app: FastAPI):
     semsimian()
     spacyner()
     # oak()
+    yield
```
glass-ships, line 25: the `on_event("startup")` listener was deprecated by FastAPI

.github/workflows/codecov_main.yaml
```diff
@@ -4,53 +4,44 @@ on:
...
         with:
           python-version: ${{ matrix.python-version }}
 
-      #----------------------------------------------
-      #          install & configure poetry
-      #----------------------------------------------
-      - name: Install Poetry
-        uses: snok/install-poetry@v1
+      - name: Install uv
+        uses: astral-sh/setup-uv@v6
```
iQuxLE, line 31: i think its best practice to pin a version of uv

.github/workflows/codecov_main.yaml
```diff
@@ -4,53 +4,44 @@ on:
...
 
-      #----------------------------------------------
-      #    install your root project, if required
-      #----------------------------------------------
```
iQuxLE, line 31: i would implement a check to verify lock with toml just to make sure both are in sync 
```
      - name: Verify lockfile is up-to-date
        run: |
          uv lock --check
```

.github/workflows/deploy-documentation.yaml
```diff
@@ -21,17 +23,19 @@ jobs:
         with:
           python-version: "3.11"
 
-      - name: Install Poetry
-        uses: snok/install-poetry@main
+      - name: Install uv
+        uses: astral-sh/setup-uv@v6
```
iQuxLE, line 28: same as above (version pinning)

iQuxLE: and would also add a lock check here

.github/workflows/codecov_main.yaml
glass-ships: probably smart, especially for now while it's still in early production.

.github/workflows/codecov_main.yaml
glass-ships: at what stage would you recommend adding this check? 
i figured `uv sync` would suffice but maybe not

backend/pyproject.toml
```diff
@@ -1,55 +1,64 @@
...
 
+[build-system]
+requires = ["hatchling", "uv-dynamic-versioning"]
+build-backend = "hatchling.build"
 
-[tool.poetry.scripts]
-monarch = "monarch_py.cli:app"
-monarch-api = { callable = "monarch_py.api.main:run" }
+[tool.hatch.version]
+source = "uv-dynamic-versioning"
```
iQuxLE, line 58: under this i think its standard to also include, just as standard configuration for it and we might want to add a fallback version

```
[tool.uv-dynamic-versioning]
vcs = "git"
style = "pep440"
fallback-version = "0.0.0"
```

.github/workflows/codecov_main.yaml
iQuxLE: i think it is safest to do it before any `uv sync`?
as `uv sync` is not checking if lock and toml are out of sync
and it does not veryify if the pyproject.toml has changed since the lock was created.
this is basically just a quaility check if someone forgets to run `uv lock `

iQuxLE:
Thanks thats great, thanks for being so fast and responsive. 
We could also add `enable-cache` when setting it up and maybe make the uv versions in an env variable on top of each workflow file? 

```

env:
  UV_VERSION: "0.7.x"

- name: Install uv
  uses: astral-sh/setup-uv@v6
  with:
    enable-cache: true
    version: $ {{ env.UV_VERSION}}
```

glass-ships:
not sure why the frontend lint test is failing here but it at least is unrelated to changes made in this PR

iQuxLE:
Yes this looks very good now! 
I'll have a look at the front end lint and see if I can find the problem otherwise it's good to merge.

Thank you @glass-ships :)

iQuxLE (approved):

PR #1131, opened 2025-06-29
adding current kg deployment date as api endpoint

iQuxLE:
### Related issues

- Closes ##1119

### Summary

- adds KG_DEPLOYMENT_DATE as env var to docker container via monarch stack v3

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -78,7 +78,7 @@ install-frontend:
 
 .PHONY: model
 model: install-backend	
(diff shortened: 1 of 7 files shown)


iQuxLE:
Frontend issue is weird. Its only warnings it should not fail this workflow.

I generated a new model.py with make model. There might have been a new version of pydantic-gen involved, not sure, but somehow the diff in deletions is huge. But seems to all work.

PR #1147, opened 2025-07-08
Fix Search Bar Issue - Adjust Break Points

### LINKED ISSUE #1146: Search bar issue when window is small
The search bar has some odd behavior when the window is small. I think we need to convert to the dropdown sooner than it currently does so the search bar has more room. See video. 

https://github.com/user-attachments/assets/a5361229-50b0-4484-94dc-3317bc000366

varun-divya:
### Related issues

- Closes #1146 

### Summary

- adds ...
- changes ...
- fixes ...
- removes ...

### Checks

### DIFF
--- a/frontend/e2e/axe.test.ts
+++ b/frontend/e2e/axe.test.ts
@@ -66,8 +66,10 @@ const checkPage =
 
     /** perform interaction on selector */
     if (selector) {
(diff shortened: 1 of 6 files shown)


sagehrke (approved):
It looks fixed to me! Thanks!

PR #1176, opened 2025-07-22
Issue 1049 filter direct associations

### LINKED ISSUE #1049: Add ability to filter for only direct associations on association tables
This is probably primarily relevant to disease to phenotype and gene to phenotype, so we could just scope it to that. 

The API already has a `direct` boolean parameter, so we just need the UI to have a switch/checkbox/option to toggle it in the UI. 

If we wanted to start down the faceting road and present the number of direct vs indirect associations, it takes a little bit of facet query shenanigans:

When viewing d2p associations for MONDO:0000824 using `facet_queries=subject:"MONDO:0000824"+OR+object:"MONDO:0000824"` gives the direct count and  `NOT (subject:"MONDO:0000824" OR object:"MONDO:0000824")` gives the indirect count. Example query as a [url](https://api-v3.monarchinitiative.org/v3/api/entity/MONDO%3A0000824/biolink%3ADiseaseToPhenotypicFeatureAssociation?traverse_orthologs=false&format=json&download=false&direct=false&facet_queries=subject%3A%22MONDO%3A0000824%22%20OR%20object%3A%22MONDO%3A0000824%22&facet_queries=NOT%20%28subject%3A%22MONDO%3A0000824%22%20OR%20object%3A%22MONDO%3A0000824%22%29&limit=3&offset=0)

varun-divya:
### Related issues

- Closes #1049 

### Summary

- Add “Direct” and “Indirect” association tabs
- Update node page theme styling
- Remove the “Include ortholog” checkbox and include orthologs by default
- Remove the “Subject” and “Predicate” columns in the Direct Associations view
- Prepend “Does NOT have” to the object label for any negated (“negated: true”) direct association

### Checks

### DIFF
--- a/frontend/src/api/associations.ts
+++ b/frontend/src/api/associations.ts
@@ -108,13 +108,12 @@ export const getDirectAssociationFacetCounts = async (nodeId = "") => {
 
   const url = `${apiUrl}/association`;
   const rawResponse = await request<any>(url, params);
-  console.log("rawResponse", rawResponse);
   // 👇 Transform the nested structure into what the UI expects
 
   const categoryFacet = rawResponse.facet_fields?.find(
     (item: any) => item.label === "category",
   );
-  console.log("categoryFacet", categoryFacet);
+
   return {
     facet_field: "category",
     facet_counts: categoryFacet?.facet_values || [],
--- a/frontend/src/api/associations.ts
+++ b/frontend/src/api/associations.ts
@@ -93,3 +93,30 @@ export const getTopAssociations = async (
   nodeId = "",
   associationCategory = "",
 ) => await getAssociations(nodeId, associationCategory, 0, 5);
+
+/** Get direct association facet counts for a node */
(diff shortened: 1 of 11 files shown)


sagehrke (changes requested):
The "All Associations" tab looks good. No comment. 
The "Direct Associations" I have a few items to look into, please:
- Can we please check that this tab is actually filtering for only Direct associations?
    - When I look at Ehlers Danlos for an example, in the D2P table, there are 82 all associations listed and then also 82 direct associations. But when I press the Phenotype Exlorer button, there are only 74 phenotypes that transfer. I am guessing there should be 82 all associations and 74 direct associations. However, if I look at FA, this issue does not appear to be a problem. 
- I like the hover for the phenotype explorer button, but i am not sure the placement make sense. With the arrow, it looks like it is pointing to the search box. I would like to see it in a different location and/or remove the arrow.

PR #1200, opened 2025-09-05
Phenotype Node Page Redesign

### LINKED ISSUE #1198: Redesign phenotype node pages on mi.org

varun-divya:
### Related issues

- Closes #1198 

Please make sure to test the edge-case phenotypes as well . 🙂


sagehrke (changes requested):
- [ ] The "Disease Phenotype" section should be called "Diseases with this Phenotype" and the tab should be "Directly associated diseases" and "Inferred associated diseases"  
- [ ] The "Causal Gene Phenotypes" section should be called "Causal Genes" and the tab should be "Directly associated causal genes" and "Inferred associated causal genes" 

I think the rest is ok! Thanks!

PR #1251, opened 2026-01-30
feat: Dynamic Open Graph meta tags for social media link previews

### LINKED ISSUE #1242: Add dynamic Open Graph meta tags for entity pages (better link previews)
## Problem

When sharing links to specific entity pages (e.g., `https://monarchinitiative.org/MONDO:0100099`) on Slack, Twitter, Discord, or other platforms, the link preview always shows the generic site description:

> "The Monarch Initiative integrates, aligns, and re-distributes cross-species gene, genotype, variant, disease, and phenotype data."

Instead of showing entity-specific information like the disease name and description.

## Root Cause

The frontend is a Single Page Application (SPA) that updates meta tags via JavaScript after page load. However, social media crawlers (Slackbot, Twitterbot, etc.) don't execute JavaScript - they only read the initial HTML response, which contains the static/generic meta tags.

### Current Implementation
- `frontend/index.html` - Has hardcoded OG tags from build-time environment variables
- `frontend/src/global/meta.ts` - Updates tags via DOM manipulation (too late for crawlers)
- `frontend/src/pages/node/PageNode.vue` - Sets node-specific description, but only after JS runs

## Proposed Solution: Backend Meta Tag Injection

Detect crawler requests at the Nginx level and proxy them to a FastAPI endpoint that serves HTML with dynamic Open Graph tags.

### Architecture

```
Request for /MONDO:0100099
         │
         ▼
      Nginx
         │
         ├─── Is it a bot/crawler? ───► Yes ──► Proxy to FastAPI /meta endpoint
         │                                              │
         │                                              ▼
         │                                      FastAPI fetches entity data,
         │                                      returns HTML with dynamic OG tags
         │
         └─── No (regular user) ──► Serve static index.html (current behavior)
```

### Implementation Steps

1. **Add FastAPI endpoint** for serving HTML with dynamic meta tags
   - New endpoint that catches entity routes (e.g., `/meta/{entity_id}`)
   - Fetches entity data from Solr
   - Injects entity name/description into OG tags
   - Returns modified HTML

2. **Update Nginx config** to detect crawlers and route appropriately
   - Use `map` directive to identify crawler user agents (Slackbot, Twitterbot, facebookexternalhit, LinkedInBot, Discordbot, Googlebot, etc.)
   - Proxy crawler requests for entity pages to the backend
   - Regular users continue to get static SPA (no performance impact)

3. **Update Docker build** to make `index.html` template available to backend

### Expected Result

| Platform | Before | After |
|----------|--------|-------|
| Slack | "The Monarch Initiative integrates..." | "COVID-19 \| Monarch Initiative - A disease caused by SARS-CoV-2..." |
| Twitter | Generic site card | Entity-specific card with name and description |

### Benefits
- Minimal changes to existing architecture
- No impact on regular user performance (static files still served)
- Only crawlers hit the backend for HTML
- Can extend later to generate dynamic OG images

## Alternatives Considered

1. **Server-Side Rendering (SSR)** - Too large an architectural change (would require Nuxt migration)
2. **Prerender service** (e.g., Rendertron) - Adds another service to maintain
3. **Static site generation** - Loses SPA benefits, complex build process

kevinschaper:
## Summary

Enables entity-specific link previews when sharing Monarch URLs on social media platforms (Slack, Twitter, Discord, Mastodon, Bluesky, etc.).

Closes #1242

### Architecture

```
Request for /MONDO:0005148
         │
         ▼
      Nginx
         │
         ├─── Is User-Agent a bot? ───► Yes ──► Proxy to /v3/api/meta/MONDO:0005148
         │                                              │
         │                                              ▼
         │                                      FastAPI fetches entity from Solr,
         │                                      returns HTML with dynamic OG tags
         │
         └─── No (regular user) ──► Serve index.html (SPA loads normally)
```

### Nginx Bot Detection

Nginx uses a `map` directive to detect crawler User-Agents and routes them to the meta endpoint:

```nginx
map $http_user_agent $is_bot {
    default 0;
    ~*Slackbot 1;
    ~*Twitterbot 1;
    ~*facebookexternalhit 1;
    ~*Discordbot 1;
    ~*Mastodon 1;
    ~*Bluesky 1;
    # ... and more
}
```

Only CURIE-format paths (e.g., `/MONDO:0005148`) are eligible - regular pages like `/about` are unaffected.

### Meta Endpoint

The FastAPI endpoint fetches the entity from Solr and renders HTML with OG tags:

```python
@router.get("/meta/{entity_id:path}", response_class=HTMLResponse)
async def get_meta_page(entity_id: str, request: Request) -> HTMLResponse:
    entity = solr().get_entity(entity_id, extra=False)
    
    # Build title and description from entity data
    title = f"{entity.name} | Monarch Initiative"
    description = f"{entity.name} - {entity.description}"
    
    # Derive URL from request headers (works for beta and prod)
    base_url = get_base_url(request)
    
    return template.render(title=title, description=description, ...)
```

### Dynamic URL Support

URLs are derived from request headers, so the same deployment works for both environments:
- `beta.monarchinitiative.org` → OG tags reference beta URLs  
- `monarchinitiative.org` → OG tags reference production URLs

## Supported Platforms

Slackbot, Twitterbot, Facebook, LinkedIn, Discord, WhatsApp, Telegram, Mastodon, Bluesky, Googlebot, Bingbot, Embedly, Pinterest, Applebot

## Files Changed

| File | Purpose |
|------|---------|
| `backend/src/monarch_py/api/meta.py` | FastAPI endpoint |
| `backend/src/monarch_py/api/templates/meta.html` | Jinja2 template |
| `services/nginx/config/default.conf` | Bot detection routing |
| `backend/tests/api/test_meta.py` | Unit tests |
| `docs/link-previews.md` | Documentation |


backend/src/monarch_py/api/meta.py
```diff
@@ -0,0 +1,100 @@
...
+        entity_id: The entity identifier (e.g., MONDO:0005148)
+        request: The FastAPI request object
+
+    Returns:
+        HTML page with entity-specific OG meta tags
+
+    Raises:
+        HTTPException: 404 if entity not found
+    """
+    try:
```
ptgolden: `SolrImplementation.get_entity` does not raise a `ValueError` when an entity is not found: it returns `None`: https://github.com/monarch-initiative/monarch-app/blob/b819149de47a51cb1930b32a97d9a234ff9d97f1/backend/src/monarch_py/implementations/solr/solr_implementation.py#L77-L93

The same logic is implemented in the `_get_entity` API endpoint: https://github.com/monarch-initiative/monarch-app/blob/b819149de47a51cb1930b32a97d9a234ff9d97f1/backend/src/monarch_py/api/entity.py#L40-L42

Also, is the blanket `except Exception` really necessary? Isn't there already something at the top level which will return a 500 error anyway?

ptgolden: (since the check for "if entity is None" is below, can all of this `try` block just be removed?)

services/nginx/tests/test_bot_detection.sh
ptgolden, line 1: These tests seem kind of silly, and they don't fit in with anything else in our testing framework. I suppose it doesn't hurt to have them in (except for the fact that we now have to maintain a pile of shell code), but I don't think they're necessary.

ptgolden (changes requested):
This looks good. My only feedback is the potential removal of that `try/catch` block and the nginx tests.

In an ideal world, we would have a slightly different architecture where this would not be necessary, and going to an entity page would just serve the meta tags properly no matter what. (i.e. the entity could be prefetched and included in the response as JSON so that the SPA could pick it up without having to immediately make a request). But that would be a much larger and more disruptive change, and it's not in the cards.

@kevinschaper you mentioned removing the `meta` endpoint from API documentation. I don't have an opinion on it either way. I don't think it hurts to leave it in, but I understand wanting to keep it "private".

services/nginx/tests/test_bot_detection.sh
kevinschaper: agree!

backend/src/monarch_py/api/meta.py
kevinschaper: thanks for catching this!

PR #1311, opened 2026-05-12
fix: left nav panel cut off on small screens

### LINKED ISSUE #1227: Left nav panel cut off on small screens
The left navigation panel on node pages is cut short on smaller screens (I am on a 13"). Maybe it needs the ability to scroll the navigation? 

https://github.com/user-attachments/assets/11009110-99d7-44b5-aea9-ff258cd345ef

varun-divya:
## Summary

Fixes the left table-of-contents (TOC) panel on node pages, which was getting cut off on smaller screens and overlapped by the footer on scroll.

- **Panel scrolls internally** when its content doesn't fit (was previously `height: 100%` with no overflow handling, leaving bottom items unreachable on short screens).
- **Panel slides up with the page when the footer enters the viewport** instead of sitting underneath the footer. The TOC's top position is offset by however much the footer has intruded from below, so the panel exits the top of the viewport as the footer takes its place — no more overlap.
- **Extra spacing between section links on short viewports** (≤900px tall) so the now-visible list isn't cramped. Taller viewports keep the original tight spacing.
- **Small bug fix**: `Math.max(header.top + header.height)` was a no-op (single-arg `Math.max` returns the value unchanged), so `nudge` could go negative when the header scrolled out of view. Now clamped to `Math.max(0, ...)`.

## Implementation

- `nudge` (header offset) and `footerOverlap` (footer intrusion into viewport) are computed in `updatePosition()` and exposed to CSS via inline `style` (`--nudge`, plus `top` derived from both).
- `.toc` uses `height: calc(100vh - var(--nudge, 0px))` and `overflow-y: auto` so the panel is always the full available height with internal scrolling when needed.
- `top` is `Math.max(0, nudge - footerOverlap)` so the panel never gets positioned above the viewport edge.

Fixes #1227

## Test plan
- [ ] In DevTools device mode at 1280×800 (or shorter), open a node page like `/node/MONDO:0019391` and confirm the left nav panel shows a scrollbar when content overflows and all bottom links are reachable.
- [ ] Scroll down on a node page until the footer enters the viewport — the TOC panel should slide up along with the rest of the page (its bottom should never go under the footer), and eventually exit the top of the viewport.
- [ ] Scroll back up — the TOC should re-pin below the header smoothly.
- [ ] At 1280×1080 (or taller), confirm there's no visible regression on full-height screens: section links are tight, panel sits below header, behavior is the same.
- [ ] Navigate between pages and confirm the panel doesn't get stuck in a partial/offset state after a route transition (verifies `footerOverlap` is reset when the footer element is briefly absent).

### DIFF
--- a/frontend/src/components/TheTableOfContents.vue
+++ b/frontend/src/components/TheTableOfContents.vue
@@ -8,11 +8,7 @@
       align-h="stretch"
       align-v="top"
       :class="['toc', { expanded }]"
-      :style="{
-        top: nudge + 'px',
-        '--nudge': nudge + 'px',
-        '--footer-overlap': footerOverlap + 'px',
-      }"
+      :style="{ top: nudge + 'px', '--nudge': nudge + 'px' }"
       role="doc-toc"
       aria-label="Page table of contents"
       @click.stop
@@ -89,8 +85,6 @@ const entries = ref<Entries>([]);
 const expanded = ref(window.innerWidth > 1240);
 /** how much to push downward to make room for header if in view */
 const nudge = ref(0);
-/** how much the footer overlaps the bottom of the viewport (so toc can shrink to avoid it) */
-const footerOverlap = ref(0);
 /** whether to only show one section at a time */
 const oneAtATime = ref(false);
 /** active (in view or selected) section */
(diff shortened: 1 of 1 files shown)


kevinschaper (approved):
thank you!

PR #1314, opened 2026-05-21
Add Open Scientist button to homepage hero tools

justaddcoffee:
Per convo with @kevinschaper @cmungall on Monarch Data call

## What

Adds an **Open Scientist** button to the homepage hero tools, alongside the existing *Phenotype Similarity Tools* and *Text Annotator* buttons. It links out to [openscientist.io](https://openscientist.io) and opens in a new tab.

## Why

Requested addition to surface Open Scientist from the Monarch front page.

## Notes

- Implemented as a third `AppLink` styled identically to the existing two buttons (`class="link"`).
- `AppLink` auto-detects the absolute URL and renders an external `<a target="_blank">`. Used `no-icon` so it visually matches the sibling buttons (no external-link arrow). Happy to drop `no-icon` if you'd prefer the external-link indicator, or swap the text for an actual logo image if there's an asset.

### DIFF
--- a/frontend/src/assets/openscientist-logo.svg
+++ b/frontend/src/assets/openscientist-logo.svg
@@ -1,0 +1,15 @@
+<svg viewBox="0 0 100 100" xmlns="http://www.w3.org/2000/svg">
+  <!-- Abstract S-shaped path representing flow of discovery -->
+  <path
+    d="M22 18 Q50 18 50 40 Q50 60 78 60 Q78 82 50 82 Q22 82 22 60"
+    fill="none"
+    stroke="#0891b2"
+    stroke-width="10"
+    stroke-linecap="round"
+  />
+
+  <!-- Three nodes: hypothesis, analysis, discovery -->
+  <circle cx="22" cy="18" r="10" fill="#06b6d4" />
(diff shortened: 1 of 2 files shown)


justaddcoffee:
Thanks for the review. Addressing the points:

- **Logo asset** — keeping it. It isn't an AI-generated placeholder; it's the first-party OpenScientist logo (`logo.svg` from the OpenScientist project), so there's no third-party branding concern.
- **`aria-hidden`** — added to the logo SVG in a2c29d4, since the button already has the "Open Scientist" text label.
- **PR description** — updated; the stale `no-icon` mention is removed (the prop isn't used — the external-link icon is already suppressed because the slot isn't plain text).
- **File location** — left at `src/assets/openscientist-logo.svg` to match the existing standalone `checkmark.svg`; moving it under `assets/icons/` would make it eligible for the `AppIcon` glob/testbed, which isn't intended here.
- **`rel="noopener noreferrer"`** — agreed it's worth fixing in `AppLink.vue`, but it's pre-existing and affects all external links site-wide, so leaving it out of this PR as a separate follow-up.

kevinschaper (approved):
looks good, sorry about the automated review's take on the logo 🤦

PR #1351, opened 2026-07-08
fix(search): format taxon labels as italic scientific names (#1350)

### LINKED ISSUE #1350: Taxon labels should be formatted correctly
Right now taxon labels in the search filter are capital case normal text, instead only the first word should be capitalized, and the whole label should be italicized. 

Image

kevinschaper:
Closes #1350.

Taxon labels were run through lodash `startCase`, which title-cased scientific names (`Homo sapiens` → `Homo Sapiens`) and mangled strain suffixes (`Saccharomyces cerevisiae S288C` → `S 288 C`). Taxon also rendered roman everywhere instead of italic.

> **Heads up: this PR is much larger than its title.** It started as a small formatting fix that turned out not to work at all. Chasing *why* uncovered a duplicated component and a second way taxon reaches node badges, and neither could be left alone without the fix staying broken or half-applied. The story is below, because the bugs are more interesting than the diff. The three commits are readable separately.

## How this got here

**1. The fix didn't work, and said it did.**
The original commit edited `TabSearch.vue`, checked that the API returns correctly-cased values, confirmed `vue-tsc` passed, and called it done. All true. All compatible with the feature being completely invisible. It even pre-explained away the one check that would have caught it, warning that local HMR "can keep rendering the pre-change options" — the exact observation that mattered.

Loaded on the PR's own branch, unchanged from the bug:

```
options = ["Xenopus Laevis", "Canis Lupus Familiaris", "Homo Sapiens", ...]
italic  = 0/11
```

**2. The page was a fork of the component we fixed.**
`TabSearch` is only ever mounted with `:minimal="true"` (from `TheHeader`), and everything the fix touched is gated behind `v-if="!minimal"` — it cannot render under any usage in the app. The live page was `PageSearchResults.vue`, a near-verbatim copy carrying its own `startCase` call. Two ~505-line files whose *entire* difference was a breadcrumb, a page title, `width="big"`, two style lines, and a help button. It even carried dead CSS for a button it didn't have.

The first instinct — patch the fix into the second copy — would have worked, and left the trap armed for the next person. So the fork went instead.

**3. Taxon reached node badges under two different prop names.**
With the search page fixed, taxon still rendered roman in gene links everywhere else. `AppNodeBadge` italicized on `in_taxon_label`, but half the callers passed the same value through the generic `info` prop, so the badge had no idea the string was a species. `AssociationsSummary` did **both**: `info:` in its template, `in_taxon_label:` in its breadcrumb helper twelve lines below.

That one only surfaced because the change was checked in a browser. The class binding rendered `class=""` and the fix was silently doing nothing.

**4. There was a third select component.**
Review caught that `infoItalic` added to `phenotype-explorer.ts` was a no-op: it imports its `Option` type from `AppSelectTags.vue`, not `AppSelectAutocomplete.vue`. TypeScript accepted the excess property and dropped it. The phenotype explorer would have shown roman taxon while the same page's header autocomplete showed italic.

**The through-line:** every one of these is the same bug. The same value living in two places, so fixing one leaves the other wrong — and each copy looks correct in isolation. The consolidation and the new page-boundary test are the only parts of this PR that make the next one less likely; the rest is cleanup.

## Changes

**`fix(search): format taxon labels as italic scientific names`** (original)
`buildFacetOption` in `util/searchFacets.ts` keeps the API's casing for `in_taxon_label` and title-cases everything else; `italic` flag on `AppSelectMulti` options, `infoItalic` on `AppSelectAutocomplete`.

**`fix(search): italicize taxon wherever a node badge shows it`**
Every badge caller now passes `in_taxon_label`. `AppNodeBadge` builds its parenthetical from parts rather than a joined string, so taxon renders italic while the id stays roman — reachable, since `ThePhenogrid` and `PageTextAnnotator` pass `show-id`. `AppSelectTags` gains the `infoItalic` flag `AppSelectAutocomplete` already had.

**`refactor(search): make PageSearchResults use TabSearch instead of forking it`**
The page drops to 23 lines of chrome; `TabSearch` owns the search box, facets, results and pagination. `TabSearch` absorbs the page's `width="big"` and the search row's max-width/margin, scoped to full-page mode so the header box is untouched. Net −495 lines.

## Reviewer notes

- **`.info { margin-left: 0.35em }` changes every badge, not just taxon ones.** Vue condenses the whitespace before the parenthetical, so `main` renders `BRAF(Homo sapiens)`. This fixes that everywhere, including `show-id` surfaces with no taxon (`HP:0001234(…)` → `HP:0001234 (…)`). Correct, but it is the change most beyond the stated scope — easy to split out.
- **The help "?" button is gone.** It lived in `TabSearch`'s dead branch and never rendered; keeping it would have made it newly *appear* on the results page. Deleting preserves today's UI. Say so if it was meant to ship.
- **`AppNodeBadge`'s `info` prop is now unreachable** — no caller passes it. Left in as a generic slot; happy to remove.
- **`omit(gene, 'in_taxon_label')`** in `SectionOverview.vue:137` / `SectionClinicalReources.vue:43` is now redundant with `:show-taxon="false"`. Not touched.
- **`SourceAssociationBrowser.vue`** has its own `badgeColumn` block that looks like another copy of the association table. Not chased; possibly the same disease.

## Testing

Verified in a browser, not just typecheck — the original PR passed `vue-tsc` while being entirely invisible.

- `/results?search=gal4` — names plain (`GAL4`), no duplicated taxon, 10/10 italic taxon badges; facets `Homo sapiens`, `Saccharomyces cerevisiae S288C`, `Aspergillus nidulans FGSC A4` correctly cased, italic 6/7 (`All` correctly excluded).
- `/HGNC:1097` — all 9 association badges render the species in its own italic span.
- Header search box unaffected by the consolidation: `max-width: none`, `margin-top: 0`, width 460px; page row 800px/32px, matching the old values.
- `unit/PageSearchResults.test.ts` is new and pins the page boundary — **where this bug lived**. Every existing test sits below it, so nothing could catch the page simply not running the code. Confirmed it fails against the forked page and passes against the fix.
- 294 unit tests pass; `vue-tsc` and `eslint` clean.

**Not verified:** the `show-id` + taxon combination (`(HGNC:1234 | Homo sapiens)`). Phenogrid and text-annotator pass `show-id`, but their nodes are phenotypes with no taxon, so I could not make it appear in a browser.

## Out of scope

Node labels in the Dismech pathograph come from dismech's authored `label` rather than the canonical term label — unrelated to search, belongs to Dismech's QC.

https://claude.ai/code/session_01BpftbLFiPqgA1xbiNBNJSW


kevinschaper:
Converting to draft, I'm still not seeing the change visible, unfortunately

PR #1359, opened 2026-07-10
feat(predicate): predicate definition + partial hierarchy explainer (#1358)

### LINKED ISSUE #1358: Make predicate definitions easily visible (with a partial hierarchy)
## Problem
Predicates appear throughout the app as bare labels — in association tables (`AppPredicateBadge`), the association-details modal, etc. — with no explanation. Users can't tell what a predicate *means* or how specific it is. This is especially confusing where several predicates coexist, e.g. the drug-indications section mixing `biolink:treats` (FDA-approved) and `biolink:treats_or_applied_or_studied_to_treat` (broader), or LOINC's `correlated_with` vs `related_to`.

## Ask
Make a predicate's **definition easily visible at the point of use**, and optionally show a **partial hierarchy** (its biolink `is_a` ancestors) so users can see how specific/general it is (e.g. `treats` → `treats_or_applied_or_studied_to_treat` → … → `related_to`).

## What already exists (reuse)
- `src/composables/use-biolink-model.ts` — loads `biolink-model.yaml` (cached 24h in localStorage) and exposes `getPredicateInfo(predicate)` returning `description`, `is_a` (immediate parent), `domain`, `range`, `inverse`, and mappings.
- `src/components/dashboard/PredicateDetailPanel.vue` — already renders that info (definition, domain/range) in the KG source dashboard.

So the definition and the immediate parent are already available; a partial hierarchy is just walking `is_a` up a few levels via repeated `getPredicateInfo` calls.

## Proposed approach
- Add an unobtrusive affordance on `AppPredicateBadge.vue` (an info icon, or make the badge hoverable/clickable) that surfaces the predicate's **definition** and a short **`is_a` ancestor chain** (2–3 levels).
- Add a `getPredicateAncestors(predicate, depth)` helper to `use-biolink-model.ts`.
- Load the biolink model **once** (module-level singleton or lazy-load on first interaction) rather than per badge — the model is ~MB, so don't parse it on every badge mount. Lazy-loading on hover/click is fine.
- Degrade gracefully when a predicate isn't found in the model.
- Reuse `PredicateDetailPanel` content/styling where it makes sense.

## Considerations / open questions
- Tooltip (definition only, cheap) vs. a click popover/modal (definition + hierarchy + mappings). Could do both: tooltip for the definition, click for the fuller explainer.
- Hierarchy depth and whether to link each ancestor.
- Where else to surface it (the association-details modal is a natural second home).

Small, self-contained, high-value clarity win; groundwork is already in place.

kevinschaper:
Closes #1358. Predicates carry an info icon that opens a modal explaining the relationship: its biolink definition, a partial `is_a` hierarchy (most-specific → general), and domain/range/inverse where present.

## Why

Association tables are full of predicates whose distinctions are invisible without looking them up. The live case: LOINC clinical measurements use both `correlated_with` (811 edges) and `related_to` (972,945), and nothing on the page says how they differ.

## How

Reuses the existing `useBiolinkModel` — `biolink-model.yaml`, cached 24h in localStorage — and adds `getPredicateAncestors(predicate, depth)` to walk `is_a`. The model loads **lazily on first open**, once rather than per badge, so a page full of predicate badges costs nothing until someone asks. Predicates absent from the model degrade to no icon.

## Open question for review: density

The icon currently appears on every predicate badge, including every association-table row. That may be too dense. Alternatives, roughly cheapest first:

- show on hover only
- restrict to section headers and the details modal
- hover tooltip with just the definition, reserving the modal for the hierarchy

Hierarchy depth is currently 3 and ancestors are not linked; both are easy to change. `getPredicateInfo` already returns exact/narrow/broad mappings, which the modal does not yet surface.

## Tests

`getPredicateAncestors` is unit-tested — chain walk, depth limit, unknown predicate — by injecting a fake model through the localStorage cache, so the tests don't depend on the shipped biolink model.

| Suite | Result |
|---|---|
| Frontend unit | 324 passed |
| Backend `pytest tests` | 607 passed, 16 skipped |
| Types / lint | clean |

Verified against a live page: 6 triggers on MONDO:0007947, modal renders the definition and the full `related to → associated with → genetically associated with → gene associated with condition` chain.

https://claude.ai/code/session_018c9UddtKsKtm35YNrxHWuL


kevinschaper:
Reviewed, and it found two ways the explainer showed the **wrong relationship**. Both are fixed on this branch, and both fixes are regression-tested against the unfixed code.

**Stale content when a table row is reused.** `AppTable` keys rows by index, so paging, sorting or filtering an association table hands `AppPredicateInfo` a new `predicate` on an instance that already has state. The lookup was cached on first open (`if (info.value) return`) and never recomputed. Open the explainer on a `treats` row, page forward to where that position now holds `causes`, open it again: the heading says `causes` while the definition, hierarchy and inverse under it still describe `treats`. Now recomputed on every open — these are dictionary hits against an already-loaded model, so there was nothing worth caching. The expanded-children toggle leaked across opens the same way and is reset too.

**Search highlighting leaked into the predicate identity.** The explainer took its predicate from the same computed the label uses, which prefers `highlighting.predicate[0]`. Solr wraps matches in markup — verified directly against the index, it returns `biolink:applied_to_<em>treat</em>`. That's right for the label, which renders as HTML, but as an identifier it matches nothing in the biolink model, so the modal claimed no definition exists, and the tags leaked into the button's accessible name and the modal title. Now taken from `association.predicate`; highlighting is display-only.

Frontend: 46 files / 326 tests pass, typecheck and lint clean.

## Not addressed, for you to decide

**The density question in the description is still open** — the icon renders on every predicate badge including every association-table row. That's a product call, so I left it alone.

**Two claims in the description are false** and should be corrected or implemented:
- "Predicates absent from the model degrade to no icon" — the icon renders for every non-empty predicate; absent ones get an icon whose modal says "No definition found".
- "domain/range/inverse where present" — only `inverse` is rendered; `domain` and `range` are read into `PredicateInfo` but displayed nowhere.

**Nested-modal accessibility.** This PR creates the first nested modal in the app: `AssociationsTable` renders `SectionAssociationDetails` in an `AppModal`, which now contains a badge with its own `AppModal`. `AppModal` wasn't designed for nesting — Escape closes both at once, and closing the inner one clears `inert`/`aria-hidden` from `#app` while the outer is still open, so the background becomes reachable by keyboard and screen readers. That's an `AppModal` limitation this PR exposes rather than causes; worth its own issue unless you want it handled here.

**Smaller, non-blocking:** `getPredicateAncestors` lists a predicate as its own ancestor under a mutual `is_a` cycle (theoretical — the shipped model is acyclic); a 200 response that parses but isn't the biolink model gets cached for 24h with no retry path; alias-resolved predicates get empty children and a broken docs link; and the hierarchy silently truncates at depth 3 with no indication the top row isn't the root.

Also confirmed: the `TS2742` error I mentioned earlier in `use-floating.ts` is an artifact of symlinking `node_modules` into a scratch worktree, not a real issue and not part of this PR. And the lazy-load claim holds — 100 badges trigger zero fetches, the first open fetches exactly once, and concurrent opens share one request.

Issue #1391, opened 2026-08-20
search: allow filtering by subset (and taxon) — VeNom veterinary terms poison grounding of human text


kevinschaper:
## Problem

`/v3/api/search` has no way to restrict results by subset (or taxon). For any caller doing
**grounding of human clinical text**, phenio's VeNom veterinary terms are a systematic source
of wrong top hits, because they are lexically excellent matches.

Concrete example — grounding the string `septic shock` against `category=biolink:Disease`:

```
#1  MONDO:1014822   septic shock, non-human animal
#2  MONDO:0017944   invasive non-typhoidal salmonellosis
#3  MONDO:0018183   staphylococcal necrotizing pneumonia
```

The `#1` hit is a veterinary term. It is returned first because it is the *only* exact lexical
match — see "not just ranking" below.

## Scale

Measured against `mondo.obo` (`data-version: releases/2026-08-04`):

| | count |
|---|---|
| MONDO terms in a `venom_*` subset | **5,045** |
| ...named `<x>, non-human animal` | 4,897 |
| ...where a **human term named `<x>` also exists** | **943** |
| ...where no human term exists | 3,954 |

The 943 are direct lexical collisions where a subset filter would immediately return the right
answer instead of the wrong one, e.g.:

```
MONDO:1013463  labyrinthitis, non-human animal        vs human MONDO:0002008
MONDO:1013231  urethral obstruction, non-human animal vs human MONDO:0001556
MONDO:1013524  isolated optic nerve hypoplasia, non-human animal vs human MONDO:0008136
```

The other 3,954 are cases where the veterinary term is the only match at all; there, the
correct behaviour for a human-text caller is to return **nothing**, which a subset filter also
achieves. Both outcomes are better than silently returning a veterinary term.

## Request

Add filtering on `subsets` to `/v3/api/search` — e.g. `subset=` / `exclude_subset=`, repeatable,
so a caller can say `exclude_subset=venom_equine&exclude_subset=venom_small_animal…` or,
preferably, exclude by prefix / pass a curated scope.

Please also consider a taxon filter, since the same "wrong organism" problem exists for callers
of other node types.

**Note for the Mondo case specifically:** `in_taxon` and `in_taxon_label` are **null on every
Mondo term** I checked, including the VeNom ones, so a taxon filter alone would not solve this.
The VeNom terms *are* cleanly identifiable by subset (`venom_equine`, `venom_exotics`,
[...]

Issue #1382, opened 2026-08-05
test_grounding_ranks_exact_synonym_first fails locally for 'hypoplastic sternum'


gingin77:
## Problem

`tests/integration/test_solr_grounding.py::test_grounding_ranks_exact_synonym_first` fails for the `hypoplastic sternum` case when Solr is running locally:

```
    def test_grounding_ranks_exact_synonym_first(text, expected_id):
        matching_results = SolrImplementation().ground_entity(text)
        assert matching_results
>       assert matching_results[0].id == expected_id
E       AssertionError: assert 'MP:0004323' == 'UPHENO:0081193'
E
E         - UPHENO:0081193
E         + MP:0004323
tests/integration/test_solr_grounding.py:66: AssertionError
```

Grounding `"hypoplastic sternum"` returns the MP term ahead of the expected UPHENO term. The other 6 cases in that test pass.

## Why it isn't caught by CI

Every test in this file is guarded by:

```python
@pytest.mark.skipif(
    condition=not SolrImplementation().solr_is_available(),
    reason="Solr is not available",
)
```

CI has no Solr service, so all 16 tests in the file are skipped — visible in any passing run's log as `tests/integration/test_solr_grounding.py ssssssssssssssss`. The failure only appears for contributors who have run `monarch solr download && monarch solr start`, which the setup guide instructs them to do. A new contributor's first full test run therefore ends red.

## Not a regression from the environment

Confirmed by running the same test against both dependency states on the same machine and same Solr index:

| state | result |
|---|---|
| before the dependency-group change (`uv sync --extra dev`) | `1 failed, 6 passed` |
| after (`uv sync`) | `1 failed, 6 passed` |

Identical, so this is not related to #1381. The assertion depends on the ranking Solr returns, which is a property of the downloaded KG index rather than the Python environment — most likely the expectation was written against an earlier KG release and relevance ordering has since shifted.

## Questions for whoever picks this up

- Is `UPHENO:0081193` still the intended ground truth for `"hypoplastic sternum"`, or has the preferred term legitimately changed?
- If ranking is expected to drift with KG releases, should this assert exact-first placement, or membership in the top N?

## Possible directions

- Update the expected ID if the KG's preferred term has genuinely changed.
- Relax the assertion to "expected ID appears in the top N results" if strict first-place ranking is too brittle across releases.
- If this ordering matters to grounding quality, treat it as a real ranking bug in `ground_entity` rather than a test-data problem.

## Related

[...]

Issue #714, opened 2024-05-22
Implement modular ingests (sup)


sagehrke:
## Super Ticket for tracking the implementation of modular ingests

**Look here for detailed tracker -->  [Modular Ingest Tracking Table 2025 Google Sheet](https://docs.google.com/spreadsheets/d/1q_QVXAmsB8aTfNXFNleGKhBn77Oj9nVzOrNQAWNrRhU/edit?gid=0#gid=0)**
_______________

**In progress**
- [ ] https://github.com/monarch-initiative/monarch-app/issues/974
- [x] https://github.com/monarch-initiative/monarch-app/issues/946
- [ ] https://github.com/monarch-initiative/monarch-app/issues/770

**Needed**
- [ ] Documentation cleanup https://github.com/monarch-initiative/monarch-ingest/issues/622

## Notes
- Need to identify QC & reporting strategy for modular ingests. 
- Need to identify next batch of modular ingests. 
     - Tech Support for next batch: @kevinschaper @AO33 @ptgolden

sagehrke:
@ptgolden Just a kind reminder that you said you would convert our modularization task into a table we can easily see status. Note, that you can do that here in markdown or if we want to use a google sheet, that would be fine too. Let me know if you need any support and thank you!!!

ptgolden:
GitHub Flavored Markdown doesn't allow creating interactive checkboxes (aka "task lists") within tables, unfortunately.

Two options. First, I could make a Google Sheet.

Second, we can manually add checkmarks into a table, e.g.

|ingest key|download yaml|transform python|transform yaml|tests|documentation|import into monarch ingest|remove code from monarch ingest|
|-|-|-|-|-|-|-|-|
|alliance_genotype|:white_check_mark:|:white_check_mark:|:white_check_mark:|:white_check_mark:|:white_check_mark:|:white_check_mark:|:white_large_square:|
|alliance_phenotype|:white_check_mark:|:white_check_mark:|:white_check_mark:|:white_check_mark:|:white_check_mark:|:white_check_mark:|:white_large_square:|

...which would require manually editing the markdown for the table.

I'd tend towards the latter, just to keep everything in the same place, but editing the markdown by hand is a bit of a pain while working on an ingest may be a bit of a pain.

sagehrke:
Let's go with a document so it's easier to make notes and updates. 

[...]

PR #1394, opened 2026-08-21
feat(search): grounding-oriented filters, exact-match mode, and named scopes (#1391, #1392)

### LINKED ISSUE #1392: search: add an exact-match / grounding mode that abstains instead of returning a best-effort hit
## Problem

`/v3/api/search` is built for **search** (maximise recall, always return the best available
hit). **Grounding** wants the opposite contract: return a match only when it is defensibly
exact, and otherwise return nothing. Today there is no way to ask for that, and no information
in the response to reconstruct it reliably.

## Evidence

Grounding 8 disease strings taken from real SemMedDB text, `category=biolink:Disease`.
Taking the **top-1 hit** gives ~50% correct:

```
Open-angle glaucoma      -> MONDO:0005338  open-angle glaucoma                CORRECT
Traumatic brain injury   -> MONDO:0858950  traumatic brain injury             CORRECT
Ovarian Carcinoma        -> MONDO:0005140  ovarian carcinoma                  CORRECT
Acute appendicitis       -> MONDO:0005649  appendicitis                       CORRECT
Septic shock             -> MONDO:1014822  septic shock, non-human animal     WRONG
Androgen excess          -> MONDO:0020542  malignant Sertoli-Leydig cell tumor of ovary  WRONG
Drug Resistant Epilepsy  -> MONDO:0009945  pyridoxine-dependent epilepsy      WRONG
Post-Operative Pain      -> MONDO:1014066  post-operative pain, non-human animal  WRONG
```

Re-running with a client-side rule of *"accept only an exact `name` or exact `exact_synonym`
match, else abstain"* turns all four errors into abstentions, with no loss on the correct four:

```
4 grounded, 4/4 correct;  4 abstained ("no confident match")
```

That is the behaviour a grounding client needs. It should not have to be reimplemented per
caller.

## Request

Add an exact-match mode to `/v3/api/search` — e.g. `match_type=exact` (vs the current default)
— which matches only against `name` / `exact_synonym` and returns an **empty result set** rather
than a best-effort hit.

Two things that would make this properly usable:

**1. Match provenance in the response.** `score` currently comes back `null` on every item, and
there is no field indicating *which* field matched or *how*. `highlighting` is populated, so the
information exists server-side. Exposing something structured — `matched_field: name |
exact_synonym | narrow_synonym | …` and `match_type: exact | prefix | fuzzy` — would let callers
implement their own precision policy without string-comparing fields client-side. The
client-side approach only works because `name`/`exact_synonym` happen to be returned, and it
degrades silently when those fields are null.

**2. Synonym scope should be selectable, and `narrow_synonym` should not count as exact.** A
`narrow_synonym` match means the query is *broader* than the term, which is exactly how
`Drug Resistant Epilepsy -> pyridoxine-dependent epilepsy` happens above. For grounding, matching
should default to `name` + `exact_synonym`.

## Minor schema note

Items carry both `synonym` (populated) and `synonyms` (always null in what I saw), plus
`exact_synonym` / `narrow_synonym` / `broad_synonym` / `related_synonym`. Having both `synonym`
and `synonyms` is an easy footgun for consumers.

## Context

Found while grounding disease mentions for a SemMedDB-derived procedure→disease ingest for the
BDC gap analysis. Related: #1391 (subset/taxon filtering) — with both, a grounding client can
call the API directly instead of wrapping it.

kevinschaper:
Closes #1392. Addresses #1391 (with a caveat about its premise, below).

Makes `/v3/api/search` usable for **grounding** — identifying which entity a piece of text names — without the caller knowing biolink category names, which ontology is which, or that veterinary subsets exist.

| Addition | What it does |
|---|---|
| `match_type=exact` | Returns an entity only when the query *names* it — whole-string, case-insensitive, on `name` or `exact_synonym` — and an empty result set otherwise. `relevance` stays the default and is unchanged. |
| `namespace` / `exclude_namespace` | The `prefix` argument `ground_entity` always had, now on `/search`. |
| `subset` / `exclude_subset` | Over the multivalued `subsets` field; a trailing `*` is a prefix match, so `exclude_subset=venom_*` needn't enumerate the five VeNom subsets. |
| `scope=` | Named bundles — `human_disease`, `human_phenotype`, `human_gene`, `rare_disease` — echoed back as the filters actually applied. An explicit parameter *overrides* its axis rather than intersecting, so `scope=human_disease&namespace=DOID` searches DOID diseases rather than nothing. |
| `facets=` | `true` = every filterable axis, uncapped; `false` = none; omitted = the two the web UI renders (unchanged for existing callers). |

`SearchResult` gains `matched_field` (`name` | `exact_synonym` | `broad_` | `narrow_` | `related_synonym`) and `match_type`, populated in *both* modes wherever the API can verify the match itself and null otherwise. `SearchResults` gains `scope`. `fl` is now `*,score`, fixing `score` coming back null on every item — it's a pseudo-field Solr omits unless asked for.

## Evidence: the 8 SemMedDB mentions from #1392

`category=biolink:Disease`:

| query | relevance top-1 | `match_type=exact` |
|---|---|---|
| Open-angle glaucoma | MONDO:0005338 ✅ | MONDO:0005338 |
| Traumatic brain injury | MONDO:0858950 ✅ | MONDO:0858950 |
| Ovarian Carcinoma | MONDO:0005140 ✅ | MONDO:0005140 |
| Acute appendicitis | MONDO:0005649 ✅ | *abstains* |
| Septic shock | MONDO:1014822 ❌ vet term | *abstains* |
| Androgen excess | MONDO:0020542 ❌ | *abstains* |
| Drug Resistant Epilepsy | MONDO:0009945 ❌ | *abstains* |
| Post-Operative Pain | MONDO:1014066 ❌ vet term | *abstains* |

**3 grounded, 3/3 correct, 5 abstained** — one fewer than #1392 predicted. `acute appendicitis` is a *narrow* synonym of MONDO:0005649, not an exact one, so it abstains under the issue's own rule; that reads as a miscount in the tally, not a disagreement.

## Blocking dependency — and a status update

The two-query shape here (Solr decides the `name_grounding` matches; a second scan hands the synonym-only remainder to `match_provenance`) exists only because the KG's Solr schema had no `exact_synonym_grounding` copy field. monarch-ingest #714 adds one.

**That field has landed.** On a local Solr running the **2026-09-02** KG, `exact_synonym → exact_synonym_grounding` is in the schema and populated on 213,146 docs, and the single filter

```
fq=(name_grounding:"<q>" OR exact_synonym_grounding:"<q>")
```

reproduces the table above exactly. The second query, the scan-row cap and the Python scope check can all go once the deployed index is confirmed to carry it — and the `[blocked on KG build]` in the title with them.

## A caveat on #1391's premise

The issue attributes the `septic shock` failure to the VeNom term being "the only exact lexical match". It isn't an exact match at all — it wins on tokenized relevance. `"septic shock"` matches **zero** entities on whole-string name-or-synonym in any category, and of **5,045** `venom_*` entities only **25** carry any `exact_synonym` — veterinary strings like `milk fever`, not bare human disease names.

So **exact mode alone defeats the VeNom problem**, including the 943 "direct lexical collisions". Subset filtering still earns its place — curated scopes, relevance-mode search, the `rare`/`ordo_disorder` families — but it isn't what fixes grounding.

Namespace matters more. Against the 2026-09-02 index:

| category | breakdown |
|---|---|
| `biolink:Disease` | MONDO 36,091 · MPATH 180 — nearly clean already |
| `biolink:PhenotypicFeature` | UPHENO 54,206 · ZP 47,514 · XPO 22,276 · **HP 20,413** · MP 15,317 · FYPO 8,653 · … |

Grounding human phenotype text against the bare category searches a pool ~88% non-human.

## Notes for review

- **Needs a merge.** 37 commits behind `main`, `mergeStateStatus: DIRTY` — conflicts in `model.py` / `model.yaml` / `solr.py` and the Solr query fixtures, which the `hl.method=unified` change touched on the same lines this PR adds `fl` to.
- The `synonym` vs `synonyms` footgun from #1392 is real but comes from the koza-produced KG schema, so it can't be fixed app-side; `match_provenance` tolerates a bare string there rather than assuming a list.
- 105 tests: 51 unit (`test_search_grounding.py`, **no Solr required** — the Solr-backed suite skips in CI, #1382), 42 integration, 12 API-layer. Backend 674 passed / 16 skipped; frontend 302 passed, types and lint clean.

https://claude.ai/code/session_01N6fjfGqaFpyjepw23qCyLv


kevinschaper:
**Holding this until a KG build carries monarch-initiative/monarch-ingest#714.**

That PR adds an `exact_synonym` → `exact_synonym_grounding` copy-field to the entity core, which is the field this one works around. Exact mode currently needs two Solr queries: `name_grounding` decides the name matches, and a second scan finds candidates that matched *some* synonym so Python can check whether it was an exact one. The scope check has to happen in Python because `synonym_grounding` copies the union `synonym` field (exact, broad, narrow and related mixed together) and raw `exact_synonym` is case-sensitive.

With `exact_synonym_grounding` in the index, that whole shape collapses to a single filter:

```
name_grounding:"q" OR exact_synonym_grounding:"q"
```

No second query, no Python scope check, no id filter folded back in — and `match_provenance` reduces to labelling which of the two clauses matched rather than deciding membership.

So merging as-is would land a workaround with a known expiry. Once a build includes the copy-field I'll simplify `_exact_search` and re-verify against the 8 SemMedDB strings, then this is ready.

Nothing else here depends on the copy-field — the namespace/subset filters, named scopes, facet switch, and the provenance fields on `SearchResult` all stand on their own. If it turns out to be useful to have those sooner, they'd split off cleanly and leave only the exact-match mode waiting.

PR #1356, opened 2026-07-09
feat(associations): LOINC clinical-measurement sections (#1349)

kevinschaper:
Closes #1349. Makes LOINC clinical measurements a first-class entity type on the node page: association sections, a hierarchy widget, an icon and colour.

Built on the flexible section matching from #1355, now merged — LOINC's edges all carry the generic `biolink:Association` category, so they cannot be split into sections by category alone.

## What you get

**Five association sections**, keyed on predicate + object category rather than association category:

| section | predicate | object |
|---|---|---|
| Correlated Phenotypes | `correlated_with` (+ positively/negatively) | PhenotypicFeature |
| Related Chemicals | `related_to` | ChemicalEntity |
| Related Anatomy | `related_to` | AnatomicalEntity |
| Related Proteins | `related_to` | Protein |
| Related Cells | `related_to` | Cell |

Each declares `match_criteria: full`, the explicit opt-in added in #1355. That matters here: under the earlier implicit rule these sections got full-criteria matching for free by having a key, which is exactly how this branch silently lost the behaviour when it fell behind its base. The same mappings drive the reverse "… Clinical Measurements" sections on phenotype/chemical/anatomy pages via `object_label`.

**A hierarchy widget** for `biolink:ClinicalMeasurement`, plus the category icon and colour.

## The hierarchy widget needed two fixes, and neither was the one I expected

The widget never needed phenio closure — `_get_node_hierarchy` is built from direct `subclass_of` edges, and the LOINC ingest does produce them. It was already working. Two things hid it:

1. **The gate was category-only**, so a term with no parents or children rendered a box containing just itself. That is an edge case for the original three categories but the norm for LOINC: only **13,627 of 107,791** terms have a `subclass_of` parent. The gate now also requires an actual hierarchy.

2. **The TOC positioned itself off-screen on short pages.** It lifted itself by the footer's intrusion into the viewport (`top: nudge - footerOverlap`); when the footer intrudes further than the panel is tall the result goes negative. Measured `y=-190` on LOINC pages against `y=109` for MONDO and HP, at every scroll position. The panel now stays pinned below the header and shrinks by the overlap instead — verified its bottom edge still lands exactly on the footer's top, both at the bottom of a long page and on a short one.

The second is why the first wasn't visible: the hierarchy was rendering correctly for LOINC terms that have parents, just above the viewport. Clinical measurements exposed it because they have few enough sections to make the TOC short — it is a general bug, not a LOINC one.

## Verified against a live index

[...]

PR #1419, opened 2026-09-02
Regenerate category enums from the KG (verified against monarch-kg-dev 2026-09-02)

kevinschaper:
`category_enums.py` is generated from the live KG and had drifted — it was last built against an older index. Regenerated against **monarch-kg-dev 2026-08-20**, the build main's fixtures already come from (#1393).

## What changed

| enum | before → after | |
|---|---|---|
| `EntityCategory` | 18 → 19 | **+ `biolink:ClinicalMeasurement`** |
| `AssociationCategory` | 27 → 27 | + `ChemicalOrDrugOrTreatmentAdverseEventAssociation`<br>− `ChemicalOrDrugOrTreatmentToDiseaseOrPhenotypicFeatureAssociation` |
| `AssociationPredicate` | 40 → 63 | +28, −5 |
| `MappingPredicate` | unchanged | |

The headline is `biolink:ClinicalMeasurement`. The KG carries **107,791** LOINC entities under it, but without the enum entry `GET /search?q=glucose&category=biolink:ClinicalMeasurement` returns **422**. With it, 1,020 hits. That unblocks searching the new entity type, which #1356 otherwise ships an icon, colour and hierarchy label for.

## Removed values — checked before committing

The five dropped predicates (`ameliorates_condition`, `contraindicated_in`, `is_active_in`, `preventative_for_condition`, `treats_or_applied_or_studied_to_treat`) have **no references** anywhere outside the enum.

The dropped association category **does** have references — `association_type_mappings.yaml` (the "Medical Action" section) and `use-association-categories.ts`. Nothing breaks: both reference it as a plain string rather than an enum member, and the full suite passes. But it makes visible a regression that arrived with the KG refresh, not with this change:

```
ChemicalOrDrugOrTreatmentToDiseaseOrPhenotypicFeatureAssociation:   0 edges
ChemicalEntityToDiseaseOrPhenotypicFeatureAssociation:             56 edges   (biolink:applied_to_treat)
```

The edges moved. So **the Medical Action section renders empty on every page that should show it** — MONDO:0015280 loses 14, several HP terms 6 each.

Repointing the mapping is a one-line change, but it's a user-visible behaviour change rather than a regeneration, so I've deliberately kept it out of this PR. Worth deciding separately whether Medical Action should follow the edges to the new category, or whether #1357's `drug_indications` section — which already spans both categories and so picks them up — is meant to supersede it.

## Verification

Regenerated by `make category-enums` against the local Solr loaded with 2026-08-20 (16,066,449 associations), not hand-edited.

| Suite | Result |
|---|---|
| Backend `pytest tests` | 592 passed, 16 skipped |
| Frontend unit | 303 passed |
| Frontend types / lint | clean |

https://claude.ai/code/session_01N6fjfGqaFpyjepw23qCyLv


kevinschaper:
intended this to be against 2026-09-01 build, but made this before I realized that build hadn't gone through

kevinschaper:
Reopened. The 2026-08-20 in the old title was the stale part, not the content.

I re-ran `make category-enums` against the live **2026-09-02** index and diffed it against what's committed here:

```
identical value set: True  (109 values)
added:   none
removed: none
```

The only difference the newer build produces is **three reordered pairs** — the generator orders by facet count, so ordering shifts with the data while the value set doesn't. I reverted that reordering rather than pushing it, since it's noise. Retitled to say what was verified.

The branch is also current with main (0 behind), so this is still mergeable as-is.

### Why it's worth landing

Without `biolink:ClinicalMeasurement` in `EntityCategory`, the KG's **107,791** LOINC entities can't be filtered in search — `/search?q=glucose&category=biolink:ClinicalMeasurement` returns **422**; with it, 1,020 hits. That's a gap on main today, independent of the LOINC stack.

It also lets #1356 drop its hand-added enum entry. That entry is currently load-bearing there precisely because this PR was closed, and a generated file is better regenerated than hand-edited.

### The one thing to decide before merging

This regeneration **removes** `ChemicalOrDrugOrTreatmentToDiseaseOrPhenotypicFeatureAssociation` from `AssociationCategory`, because it has 0 edges in the KG. The "Medical Action" mapping in `association_type_mappings.yaml` still references it, as does `use-association-categories.ts`.

Nothing breaks — both reference it as a plain string rather than an enum member, and the full suite passes — but it leaves a dangling reference. #1357's `drug_indications` section already spans both that category and `ChemicalEntityToDiseaseOrPhenotypicFeatureAssociation` (where the 56 edges actually live), so it may simply supersede Medical Action rather than needing a separate fix.

